PR #9, opened 2025-11-02
feat: add exegol extension

lap1nou:
Greetings,

This PR add an extensions to manage Exegol containers: https://exegol.com/.

Here is the feature for now:
- Create a container based on a profile
- Open workspace folder
- Directly open a shell inside the container

Regards.

### DIFF
--- a/extensions/exegol/src/utils/cli.tsx
+++ b/extensions/exegol/src/utils/cli.tsx
@@ -4,7 +4,7 @@ import { ExegolContainer } from "../models/ExegolContainer";
 import { readFile, writeFile } from "fs/promises";
 import { join } from "path";
 import { homedir } from "os";
-import { getPreferenceValues } from "@vicinae/api";
+import { getPreferenceValues, showToast, Toast } from "@vicinae/api";
 import { existsSync } from "fs";
 import { DEFAULT_CONFIG } from "./config";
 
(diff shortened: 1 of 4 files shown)


extensions/exegol/src/utils/cli.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { readFile, writeFile } from "fs/promises";
+import { join } from "path";
+import { homedir } from "os";
+import { getPreferenceValues } from "@vicinae/api";
+import { existsSync } from "fs";
+import { DEFAULT_CONFIG } from "./config";
+
+const execPromise = promisify(exec);
+
+const EXEGOL_PROFILES_PATH = "/.exegol/profiles.json";
```
aurelleb: make sure this exists, or handle it gracefully. Right now it just shows as empty 
image

extensions/exegol/src/utils/cli.tsx
```diff
@@ -0,0 +1,125 @@
...
+	name: String,
+	profileNewContainer: String,
+	jsonProfileData: any,
+) {
+	if (profileNewContainer && jsonProfileData) {
+		var cmdFromProfile = constructCommandFromExegolProfiles(
+			jsonProfileData,
+			profileNewContainer,
+		);
+		var startCmd = `exegol start ${cmdFromProfile.join("")} ${name} ${jsonProfileData[profileNewContainer]["image"]}`;
```
aurelleb: you should check if `exegol` is installed and if not, show an helpful error message to the user upon starting the extension to let them know how they can install it

extensions/exegol/src/utils/cli.tsx
```diff
@@ -0,0 +1,125 @@
...
+		console.error("Error running Exegol containers:", error);
+	}
+
+	return;
+}
+
+export async function listExegolContainers(): Promise<ExegolContainer[]> {
+	try {
+		const { stdout } = await execPromise(
+			"docker ps --all --format '{{.Names}}' --filter \"name=^exegol-\" | cut -d '-' -f 2",
```
aurelleb: same, you should check whether docker is installed and inform the user that they need to install it if not,

aurelleb (changes requested):
Cool extension! However, it lacks basic error handling or in-extension guidance on how to use it, which makes it hard to use as an outsider.

extensions/exegol/src/utils/cli.tsx
aurelleb: image

PR #15, opened 2025-11-05
[SWWW switcher] Add support for post script to be runned after wallpaper is set,

C4n4m1:
Script is provided by user in the setting as optional feature.
Useful for custom workflow ( for example :  color generation like `matugen`, or `lockscreen` like `hyprlock` , ... )
trying to match the post script feature of `Waypaper` 

*Unfortunately I have some issues about running dev  extension with  my vicinae installation.*

### DIFF
--- a/extensions/swww-switcher/package.json
+++ b/extensions/swww-switcher/package.json
@@ -190,6 +190,14 @@
       "default": "90"
     },
     {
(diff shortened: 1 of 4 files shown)


C4n4m1:
I totally understand,  it's better this way.

PR #22, opened 2025-11-07
 extension(silverbullet): add Silverbullet extension for managing notes

knoopx:
image
image


aurelleb:
Please remove the `ray` scripts for future extensions, as they are not even used here

PR #30, opened 2025-11-09
feat: add playback-management extension

marcinjahn:
Adds simple playerctl based commands to resume, pause, play next, play previous.

The next/previous icons are not that great, but surprisingly https://ray.so/icon did not have any better icons.


aurelleb:
Waiting for @dagimg-dot input here but I guess we could also have both and let the 'market' decide which one is more popular, and maybe discontinue one of the two.

The global store is controlled because it's a curated list of extensions people might want to install. It is still possible to pass extensions around using other means, although it doesn't benefit from the same discoverability.

marcinjahn:
> but I guess we could also have both and let the 'market' decide which one is more popular, and maybe discontinue one of the two.

Good ol’ free market :smile:

PR #32, opened 2025-11-10
feat: Add more commands to player-pilot

marcinjahn:
Adds resume, pause, next, previous.
This is alternative PR after discussion in https://github.com/vicinaehq/extensions/pull/30

image


dagimg-dot:
> @dagimg-dot is it okay for you?

yeah it good. I will match the icons some other time. but functionaltiy wise its done

PR #34, opened 2025-11-10
feat: add pass extension for managing standard unix password manager entries

Tinkerbells:


aurelleb:
Sorry for the delay, I will get to it soon.

PR #40, opened 2025-11-14
feat(wifi-commander): add iwctl capabilities [WIP]

Gareck-OG:
# Add `iwctl` Support (WIP)

This PR introduces optional support for `iwctl`/`iwd` alongside the existing `nmcli` integration.  
It also includes a refactor of parts of the extension to accommodate multiple network backends.

## Status: Work in Progress
- Implemented and lightly tested:  
  - Toggle Wi-Fi on/off  
  - Restart Wi-Fi  
- Further testing is planned.

## Help Needed
Since I don’t have `nmcli` available on my system, it would be very helpful if someone could verify that the `nmcli` functionality still works as expected after the refactor.

## Question
When merging, will the commits be squashed?  
Some commits are simply “sync commits” made while switching between devices, so they're not meaningful as standalone commits.

---

Feedback is very welcome!

### DIFF
--- a/extensions/wifi-commander/README.md
+++ b/extensions/wifi-commander/README.md
@@ -12,6 +12,7 @@ A comprehensive Wi‑Fi management extension for Vicinae that provides complete 
 ### Network Scanning
 - **Scan Available Networks**: Discover all nearby Wi‑Fi networks
 - **Smart Connection**: Automatically detects saved networks and connects without password prompt
+- **Hidden Connection**: Gives to option to connect to hidden Networks [just `iwclt`]
 - **Password Protection**: Secure password form for new networks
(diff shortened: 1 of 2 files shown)


Gareck-OG:
@dagimg-dot I noticed that the current code assumes network names/SSIDs will never contain a pattern like `"NetworkName<space>continueName"`. While most users won’t name their networks this way, I’ve seen a few cases where it happens, and this could lead to errors.

I can prepare an additional PR to handle this edge case once I’m done with my current changes, if you’d like.

aurelleb:
damn, this extension is getting serious

aurelleb:
works very well with iwctl, congratulations!
For a future update, maybe you should render less details in the list because it ends up very cluttered:
image

PR #42, opened 2025-11-15
feat: omarchy menu

codingcodax:
Extension to mimic the `omarchy-menu`.

> [!NOTE]
> Several things are missing, like error handling in case the extension is installed in a non-omarchy machine.

This version is still very rough, but it works (what I tested).

All feedback is welcome

---

https://github.com/user-attachments/assets/a252b379-75c4-479c-8c99-2d8f34beaa65


aurelleb:
one nitpick, I think it's worth using this logo instead of that generic search one, unless you think it's too branded
extension_icon

extensions/omarchy-menu/src/helpers/actions.ts
```diff
@@ -0,0 +1,35 @@
+export const terminal = (command: string) =>
+  `xdg-terminal-exec --app-id=com.omarchy.Omarchy ${command}`;
```
leiserfg: Shouldn't this use `runInTerminal`?

codingcodax: Hey, I named the functions that way to match how it’s done in [omarchy-menu](https://github.com/basecamp/omarchy/blob/8350b84eee10601efce64efef6fe044f03bfcd32/bin/omarchy-menu#L39-L41), I was trying to keep things consistent. Is that the issue or is there another thing I’m missing?

aurelleb: @leiserfg I think it makes sense to use the same omarchy tooling here to make sure this remains consistent with the way they do things.
@codingcodax he is referring to the `runInTerminal` API we introduced recently: https://api-reference.vicinae.com/runInTerminal.html
You could technically implement the same behavior using this.

abelfubu:
As promised, I opened a [PR](https://github.com/codingcodax/extensions/pull/1) fixing the issues I mentioned above. Now all captures (screenshots, screen recording) work fine on my end, by closing the vicinae panel and executing the command.

Before there was a tricky issue where the panel would close but the command would not execute until the next time you toggled open the vicinae panel, this is fixed why spawning the command in detached mode.

Just let me know if this is the preferred way or you want me to make some changes :pray:

aurelleb:
you're calling it an OS? Twitter will be very angry at you.

eduwass:
@codingcodax @aurelleb omarchy quattro came out, wondering if this is up to date with it or it would require a review? ty!

aurelleb:
@eduwass probably not up to date, will require an update I think

codingcodax:
I haven't touched Omarchy for a few months, so I won't be updating it in the short term. I need to find time and see what's new.

eduwass:
all good @aurelleb @codingcodax  understandable, I just wanted to check

PR #43, opened 2025-11-15
Add ssh extension

leiserfg:


aurelleb:
works great!
I think you should add an action to clear the history, though.

PR #52, opened 2025-11-17
feat(systemd): ✨ Add systemd service management extension

knoopx:
**tested only on my nixos, needs to be tested on other os to ensure correct functionality**

image


Keyruu:
Awesome stuff man!!!!! Works on my machine, but I'm also on NixOS.

PR #54, opened 2025-11-20
feat: add power-profile manager extension

BOTKooper:
Uses `powerprofilesctl` from `power-profiles-daemon` to actually do stuff

tested on my arch & omarchy machines

it WILL break if users install the daemon on conflicting distros, not sure how to deal with that ([link to arch wiki](https://wiki.archlinux.org/title/CPU_frequency_scaling#power-profiles-daemon)

icon yoinked from somewhat similar raycast extension - [link](https://www.raycast.com/sasivarnan/power-management)

When all goes good:
image

When deps are missing
image

### DIFF
--- a/extensions/power-profile/src/power-profile.tsx
+++ b/extensions/power-profile/src/power-profile.tsx
@@ -6,7 +6,7 @@ import { ProfileSelector } from "./components/profile-selector";
 export default function Extension() {
 	return (
 		<IfCliCommandSucceeds
-			command="powerprofilesctl123"
+			command="powerprofilesctl"
 			onSuccess={<ProfileSelector />}
 			onFail={
 				<Detail


BOTKooper:
@aurelleb that sounds about right, but i'm not sure if/how we can do that without breaking the existing distro components, 

making it a question - "what would happen to {insert_distro_name} included power management thing-y if we manually change governor settings from under it (or, even worse, use non-conforming ones)"

Maybe a better long-term solution is to have a support for all daemons listed on the [wiki](https://wiki.archlinux.org/title/CPU_frequency_scaling#Userspace_tools) and use the one present? In this case the error state would be `"no known power management daemon installed - here's the list of supported options, figure it out"`

PR #46, opened 2025-11-16
Add resolved and integrity to extensions automatically

leiserfg:
This is a WIP, let's see if that fixes it.

PR #55, opened 2025-11-20
fix(nix): update package-lock.json

0xda157:
fixes https://github.com/vicinaehq/extensions/pull/50


0xda157:
> > Just noticed that error, does that mean #46 isn't working?
> 
> yes, it seems like it isn't working

hmm strange is just worked [here](https://github.com/vicinaehq/extensions/commit/9c5de422bad0679cb715b4ce124f2e7ba4c51bc4)

PR #60, opened 2025-11-22
FEAT: Omarchy update@3.2.0

abelfubu:
# :fire: Changes introduced in this pull request

- Bluetooth commands have changed in this last version, they have been updated and they work for v.3.2.0
- Screen Record commands have also changed and have been updated accordingly
- Added PHP developer install commands that were missing

> [!NOTE]
> I have been thinking in a way to handle different versions at the same time

Command could be a map 
```ts
    {
      id: "omarchy",
      name: "Omarchy",
      icon: "󰣇",
      command: {
        default: present_terminal("omarchy-update"),
        '3.2.0': 'another-command'
      },
    },
```

And be could get the version on extension launch with `omarchy-version` and match the command like:

```ts
  const { isLoading, error, data } = useExec(
    "/bin/sh",
    ["-c", "command -v omarchy-version"],
    {
      execute: true,
    },
  );
  
  spawn(item.command[data] ?? item.command.default, ...)
```

Let me know what you think and if you think this is worth it or it could get really messy really fast :pray:

### DIFF
--- a/extensions/omarchy-menu/src/config/install.ts
+++ b/extensions/omarchy-menu/src/config/install.ts
@@ -202,7 +202,26 @@ export const installMenu: MenuItem = {
           id: "php",
           name: "PHP",
           icon: "",
-          command: "",
+          items: [
+            {
+              id: "php",
+              name: "PHP",
+              icon: "",
+              command: present_terminal("omarchy-install-dev-env php"),
+            },
+            {
+              id: "laravel",
+              name: "Laravel",
+              icon: "",
+              command: present_terminal("omarchy-install-dev-env laravel"),
+            },
+            {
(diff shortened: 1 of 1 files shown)


Gareck-OG:
ok. i will open a PR on ur fork. so all 3.2 updates are together.

Gareck-OG:
I had a problem with the image rendering if i remember it correctly. image had a squashed width and a hight where scrolling was needed for the `list.item.detail` markdown. If i understood i correctly using grid instead of list would fix that issue? But didnt have time to look into it.

aurelleb:
@Gareck-OG might be a rendering problem in the markdown renderer, that's worth reporting

PR #61, opened 2025-11-22
nix: package extension that use node-gyp

0xda157:


knoopx:
fyi: the nix extension does not need `node-gyp` at all, not sure why it is being pulled

PR #63, opened 2025-11-24
fix(wifi-commander): information vomit

Gareck-OG:
I’ve cleaned up the information shown in the list.

## Question
I’m unsure whether I should remove the “connected” and “disconnected” labels in **Manage Saved Networks**.

The labels provide useful context directly in the list, but this information could also live in the detail view. Since the green checkmark already indicates whether a network is currently in use, the labels might be redundant.

What do you think?

## Screenshot Reference
image


Gareck-OG:
> Yeah it should be removed.
> 
> And also on a related note, main action on saved networks should be connect not forget.

That onlz makes sense for networks that can get conencted to. So if a network is out of range and not detected with a scan, connect option is useless no?

dagimg-dot:
> Ok. I take that back. today i accidently removed a network from my saved networks accidentally.

Told you 😁

PR #64, opened 2025-11-26
feat(screen-recording): add screen recording util

BOTKooper:
Uses `wf-recorder` to actually do stuff, so needs to be run on wlroots-based compositor

Needs pipewire to do audio, but i think pretty much anybody has it at this point - not even listing in error msg

Only _nasty_ thing is that camera overlay gets spawned as a window, so users need to set it to floating (and use smth like [magic workspace](https://wiki.hypr.land/Configuring/Uncommon-tips--tricks/#minimize-windows-using-special-workspaces)) in their WM config manually

image
image
image

### DIFF
--- a/extensions/screen-recorder/src/components/cli-check.tsx
+++ b/extensions/screen-recorder/src/components/cli-check.tsx
@@ -26,7 +26,7 @@ export const IfCliCommandSucceeds = ({
 			.finally(() => {
 				setIsLoading(false);
 			});
-	}, []);
+	}, [command]);
 
 	if (isLoading) {
 		return <List isLoading={true} />;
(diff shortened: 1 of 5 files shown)


aurelleb (changes requested):
Really cool extension concept, I would really like to be able to use that.
However it seems to silently fail on my system right now, which is not a very good experience.
See video below (recorded with obs to avoid conflict with wf-recorder. on an interesting note, when recording with wf-recorder beforehand it will show as recording inside the extension and stopping the recording using the extension will correctly shutdown the instance)

also it would be great to have an option to directy record a region using slurp or something.

Also for reference, I'm running Hyprland


https://github.com/user-attachments/assets/d1218432-22ce-4b0e-b992-d69740911815

PR #66, opened 2025-11-27
feat: fallback to swww

Magniquick:
the officall awww docs repology still shows awww being packaged as swww in a lot of distros including arch.


0x4c756e61:
I wonder if this is because it didn't have a release since its rename (if am not mistaken), regardless, yeah that's fine with me, if the binary is named swww pretty much everywhere then we should probably even default to it?

To my knowledge, the only place where the binary is named "awww" is in the nix flake or if you're compiling manually (which includes the AUR aww-git package)

PR #67, opened 2025-11-28
fix(wifi-commander): QOL

Gareck-OG:
* fix: shortcuts for manage saved networks
* fix: nmcli commands for network names with spaces
* fix information vomit


aurelleb:
much better, thanks!

PR #70, opened 2025-12-02
add vscode-recents extension

ShyAssassin:
This PR adds a new and pretty simple extensions that can be used to quickly jump back into any vscode project you have recently had open.

~~Currently there are a few limitations mainly that this only works on "normal" vscode and that you can not remove recents but i believe these are minor enough to be fixed in a follow up PR at a later date.~~
Support for other vscode forks (vscodium, cursor) has been added

image


aurelleb (changes requested):
I think only supporting vscode is a big deal because a lot more people are using various vscode forks nowdays. I would like for it to work out of the box with all the vscode forks, as long as they do not deviate from the way the recent projects are stored
In order to publish this I would like for at least vscodium and cursor to be supported

ShyAssassin:
I can pretty easily add support for vscodium but i am not 100% sure about cursor, i will quickly implement vscodium and then look into cursor

ShyAssassin:
I would like to add the git and ssh stuff to this extension but i am not exactly sure how to find that info from the DB, i will look into it but i think this extension is now at a point where is is completely usable and extensible.

PR #71, opened 2025-12-04
feat: add extension spongebob-text-transformer

Bassadin:
A fun Vicinae extension that transforms text into mocking SpongeBob case (aLtErNaTiNg CaSeS) for creating sarcastic or humorous text variations.

If the spongebob part is too off-tone for the extension store, I can change it to a more general text transformation name. I just thought i'd be the association most often used.


aurelleb (changes requested):
it's fun, I don't take any issue with that. However, it looks like you forgot to commit the assets folder, so the icons don't load.

aurelleb (changes requested):
it's fun, I don't take any issue with that. However, it looks like you forgot to commit the assets folder, so the icons don't load.

PR #73, opened 2025-12-05
Feature/omarchy theme preview

abelfubu:
# :fire:  Omarchy Theme Preview

I just found a bit of time to implement the theme preview feature @Gareck-OG 

https://github.com/user-attachments/assets/c927be6e-d041-472d-aa66-88e8d8e4fd46

It does replicate the behavior of the current menu, shows the theme preview image or the first background image of the theme as a fallback


Gareck-OG:
Weird. I didn't look into it further but everything I tried made it render weird. 

Is height 320 always going to be the height on every machine?

Gareck-OG:
Np. Just annoying that we both wrote the code to implement the preview xd. i just focused on fixing the route cause of the problems xd.

Gareck-OG:
It git merged.

PR #75, opened 2025-12-05
add extension: project-launcher

benzipperer:
This extension adds a couple of utilities to create and open folders in common editors: currently vscode, positron, cursor.

You can create a new folder and open it in your preferred editor, or you can search and open recent projects/folders.


extensions/project-launcher/package.json
```diff
@@ -0,0 +1,169 @@
...
+          "value": "code"
+        },
+        {
+          "title": "Cursor",
+          "value": "cursor"
+        }
+      ]
+    },
+    {
+      "type": "dropdown",
```
aurelleb: We have the `runInTerminal` API for this: https://api-reference.vicinae.com/runInTerminal.html
You should use this instead, and remove the preference

extensions/project-launcher/package.json
```diff
@@ -0,0 +1,169 @@
+{
+  "$schema": "https://raw.githubusercontent.com/vicinaehq/vicinae/refs/heads/main/extra/schemas/extension.json",
+  "name": "project-launcher",
```
aurelleb: this name is too generic, as it only handles a very small subset of IDEs. Consider making it more explicit.

PR #80, opened 2025-12-09
Added bookmarks extension

ShaderCompilation:
It's a simple extension that allows users to bookmark links, and later open/copy them.
There is no external dependencies/fetching, etc...
It's a simple list of links.
More info in the README.md file


aurelleb:
The new name is great, thank you for the extension :)

PR #86, opened 2025-12-15
feat(extensions): introduce Claude chat extension

dfseifert:
First of all, thanks a lot for this project!
I have been daily driving vicinae for a while now and just love it!

So I wanted to give back with creating a little extension I find quite useful.

This PR adds a new Claude chat extension for Vicinae. 

It enables AI-powered assistance through Anthropic's API, including features like persistent chat history, multiple model support, and real-time streaming.

> **Note:** This is an early example of a chat integration built with Vicinae extensions. It demonstrates how to build conversational AI experiences using Vicinae's native architecture.
> I am happy for any feedback!

image

image


dfseifert:
Hi @aurelleb 

thanks for the feedback.

I've applied a patch that fixes the following:
The default view for the new chat command is now the chat history so that when navigating back, the chat history is shown.
Also when an Error during response collection occurs, it creates a new conversation item that displays the full error message. This could then be debugged. (Or is there a better way to do this?)
Please let me know, happy for feedback!

Kind Regards
Daniel

aurelleb:
> The default view for the new chat command is now the chat history so that when navigating back, the chat history is shown

I think this is weird behavior because it pushes a view on the stack without the user doing anything which makes the back navigation feel misleading. The average user will be confused about this.
You probably should have a command to start a new chat and a separate one that allows you to browse the existing threads.


https://github.com/user-attachments/assets/269e75fe-4bf6-4885-96bc-0d362fc6457d

aurelleb (changes requested):
see comment above

PR #88, opened 2025-12-19
use a root command for each gnome setting panel

pomdtr:
Instead of beeing grouped in a sublist, this PR allows you to open gnome settings panel from the root list of vicinae.

This match raycast behavior on macOS, where you can browse preferences tab from the root list

image


aurelleb:
going forward with this change as it makes more sense.

PR #98, opened 2025-12-24
fix(pass): fixes symlinked password entries

FredrikMWold:
The `dirent.isFile()` always returns falls when the file is a symlink.
This pr adds logic to handle that case


aurelleb:
good catch, thanks!

PR #99, opened 2025-12-25
fix(hypr-keybinds): use hyprctl for keybinds instead of config file

FredrikMWold:
This PR changes the keybinding loading mechanism from using a user-defined file to querying hyprctl for all active keybindings.

This makes it so users no longer need to configure or maintain a separate keybindings file.


aurelleb:
much easier to use that way, thanks. 
Idea for a next update: make the shortcut searchable by its command (searching 'vicinae') would bring up the vicinae shortcut if there is one.

PR #105, opened 2025-12-28
feat: add aria2-manager extension

ozer-01:
This PR adds a new extension called **Aria2 Manager**.

It provides a GUI interface within Vicinae to manage `aria2c` downloads without needing to use the terminal. It supports standard HTTP downloads, Magnet links, and video downloads via `yt-dlp` integration.

## Features

- **Quick Download:** Add downloads directly from the clipboard.
- **Smart Input:** Detects URLs (HTTP, Magnet, YouTube) automatically.
- **Task Control:** Pause, Resume, and Remove active tasks.
- **File Management:** Option to delete the actual file from the disk when removing a task.

### DIFF
--- a/extensions/aria2-manager/README.md
+++ b/extensions/aria2-manager/README.md
@@ -14,7 +14,7 @@ Control your downloads directly from Vicinae using Aria2 and yt-dlp. This extens
 
 ## Features
 
-* **Quick Download:** Add downloads directly from your system clipboard.
+* **Quick Download:** Fast way to add a download. You can pass a URL as an argument, or if left empty, it will automatically read from your clipboard.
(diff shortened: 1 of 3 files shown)


aurelleb (changes requested):
the quick download command is a bit misleading because it is not clear that it expects to read the url from the clipboard. you should probably  make the error message clearer or even better, allow the command to take an optional argument (use the `arguments` field in the manifest) and fallback on clipboard if the argument is not there

ozer-01:
thanks for the feedback I just pushed the updates. The `quick-add` command now accepts an optional `url` argument and prioritizes it over the clipboard. I also improved the error messages to be more specific and actionable instead of generic errors.

Ekran Görüntüsü_20251229_002901 pn111111

aurelleb:
thanks!

PR #106, opened 2025-12-28
feat(vscode-recents): add remote workspaces and window preferences

JacobDeuchert:
Thants for the great work. With my own workflow I found some issues with the vscode-recents extension.

This PR  fixes the following issues:
- Remote workspaces / folders not showing
- Sorting not working when no lastOpened property is present in vscode state

This PR adds:
- Preference to configure whether the workspace / folder should be opened in a new window, should reuse existing window or should be default (let vscode decide)

The code (especially the parsing of the remote authority) is inspired by MS PowerToys.

I only tested my own use case with normal vscode and dev containers. Maybe I find time to test codium / cursor with other remote targets like ssh and codespaces.

image

### DIFF
--- a/extensions/vscode-recents/src/helpers.ts
+++ b/extensions/vscode-recents/src/helpers.ts
@@ -33,9 +33,7 @@ export function decodeFileUri(uri: string): string {
 }
 
 export function getProjectLabel(projectType: ProjectType, projectPath: string, projectLabel: string | undefined = undefined): string {
-    const usedPath = projectLabel ?? projectPath;
-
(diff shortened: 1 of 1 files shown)


aurelleb:
works fine for me, @ShyAssassin  can you make sure it works fine for you too?

ShyAssassin:
LGTM but i do think something should be added to more easily tell if something is a dev container or a project from another machine (ssh), would probably be worth changing the icon of the project list entry depending on if it is local or not.
Or maybe just appended the machines name to the project name?
image

ShyAssassin (approved):

PR #107, opened 2025-12-29
Add Mise Extension

FredrikMWold:
### This PR adds a new extension called Mise.

It provides a GUI interface within Vicinae to manage mise tools without needing to use the terminal. It supports browsing the mise registry, installing specific versions, and switching between installed versions.

### Features
- Tool Overview: View all installed tools with version badges and active status.
- Install: Install tools from the mise registry.
- Version Picker: Browse and install specific versions of any tool.
- Version Switching: Activate different installed versions instantly.
- Uninstall: Remove versions you no longer need with confirmation.

image

### DIFF
--- a/extensions/hypr-keybinds/README.md
+++ b/extensions/hypr-keybinds/README.md
@@ -8,7 +8,7 @@ Look at what your binds are in a hyprland config file.
 
 ## Features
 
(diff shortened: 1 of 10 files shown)


aurelleb:
great one, thanks!

PR #119, opened 2026-01-11
fix(omarchy-menu): extention crashing after theme folder location change in omarchy

FredrikMWold:
This PR fixes an issue introduced after the Omarchy 3.3.0 update.
The update changed the location of the themes directory, which caused the extension to crash on launch.


aurelleb:
thanks!

PR #131, opened 2026-01-19
feat: adds a simple Hyprland monitor extension

FredrikMWold:
New extension for managing Hyprland monitor configurations through Vicinae. 

Detects connected monitors and allows changing resolution, refresh rate, scaling, rotation, and positioning via a form interface. Includes a 10-second safety countdown to confirm changes or auto-revert. 

Configurations are persisted to vicinae-monitors.conf which is automatically sourced in the main Hyprland config.


extensions/hyprland-monitors/src/MonitorSettings.tsx
```diff
@@ -0,0 +1,164 @@
...
+    >
+      <Form.Description text={monitor.description} title="Monitor name" />
+      <Form.Description text={monitor.name} title="Port" />
+      <Form.Dropdown
+        id="resolution"
+        title="Resolution"
+        onChange={setMode}
+        value={mode}
+      >
+        {monitor.availableModes.map((mode) => {
```
aurelleb: you should dedupe the resolutions, for some reason it is possible under some setups for the same resolution to appear many times.

aurelleb (changes requested):
I agree with @fufexan here that modifying the Hyprland config as a default behavior is a bit too intrusive imo. It should probably be a secondary action on the apply changes view, "apply and save". And you could add a preference to decide which one should be the default, so that the user can tweak that.

Also, most of the other comments are probably weird edge cases stemming from the fact I'm using a single display split in two physical displays using PBP. But nevertheless, it should be handled correctly.

FredrikMWold:
@aurelleb I did not mean to close the PR.
I have fixed the things mentioned

PR #132, opened 2026-01-20
nix: don't unnecessarily use an ifd

0xda157:
IFDs are highly discouraged and there's no reason to use one here (now that https://github.com/vicinaehq/vicinae/pull/979 has been merged)


aurelleb:
thanks!

PR #144, opened 2026-02-02
feat: add 'color converter' extention

Costeer:
In essence, it's a tool for developers who frequently work with colors and need to quickly convert them between different representations or find their Tailwind CSS equivalents.

image

### Features
If the input is valid, it converts the color into the following formats:
- HEX
- RGB
- HSL
- CMYK
- HTML color name
- Tailwind color name

### Disclaimer
the extention is 100% written by Ai as i am not a react dev, i looked over the code and validated that everything lines up with the guidelines


extensions/color-converter/.direnv/bin/nix-direnv-reload
aurelleb: all the files under .direnv should not be commited, similarly the rest of the nix stuff

extensions/color-converter/flake.nix
aurelleb: do not commit

PR #154, opened 2026-02-10
ProtonVPN extension

Mujtaba1i:
This extension adds a new Vicinae extension that enables quick command-based control of ProtonVPN directly from Vicinae.

Features:

- Connect to ProtonVPN
- Disconnect from ProtonVPN
- Toggle VPN connection state
- View current VPN status
Repository:
https://github.com/Mujtaba1i/vicinae-protonvpn-extension

The extension builds successfully using the Vicinae SDK and is ready for store inclusion.


magnushinge:
Hey Mujiba.
I have 3 things I think you should add to this to make it more user friendly:
1. Add a screen that says you are missing the dependency proton-vpn-cli if it is not installed on the user's machine
2. If you are not logged in and try to use any of the commands, open a menu where you can log in, or if that is too hard since the CLI seems like a pain to work with, just a screen explaining that you are not logged in and how you can log in. Currently it just throws a "Command failed" which can easily confuse users
3. Add a split view in the "Dashboard" where you can see your network status instead of it being an independent menu, I think it makes more sense to just have it in the dashboard. 
Maybe take inspiration from how the bluetooth plugin did this:

Screenshot from 2026-03-19 23-47-21

PR #158, opened 2026-02-12
fix(nix): respect keyboard common bindings config

### LINKED ISSUE #157: Extensions must respect vicinae keybinds config for generic Action shortcut
Many extension have defined <Action> and have added hard key mapped shortcuts.

Vicinae config has generic for "Copy action" and more.
All extension must respect that first, and then fallback to default.

Eg notice in nix extension

https://github.com/vicinaehq/extensions/blob/216d366f8fc08b3eb427ebffa25dce14c3c6ed7f/extensions/nix/src/components.tsx#L36-L56

Here "Copy option name" is bound to Ctrl+n, this breaks emacs mode of keybindings and Imo C-n does not make sense for copy.

Quick grep for `shortcut={{` shows many hard coded bindings. Everything can be made sure to use generic user key maps?

Image

idlip:
Partially closes #157 

- Maps all shortcuts to Keyboard.Shortcut.Common

I've used the names from 
https://github.com/vicinaehq/vicinae/blob/10c98516b63bbb790a63b0ed84724352f32bc08d/src/typescript/api/src/api/keyboard.ts#L108-L125

Please LMK if this needs to be done for all other packages.

I've used Copy, CopyName and CopyPath for 3 copy shortcuts here.

@knoopx Please LMK if the changes are fine?

### DIFF
--- a/extensions/nix/src/components.tsx
+++ b/extensions/nix/src/components.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { exec } from "child_process";
-import { Action, ActionPanel, Color, Detail, getPreferenceValues, Icon, List } from "@vicinae/api";
+import { Action, ActionPanel, Color, Detail, getPreferenceValues, Icon, Keyboard, List } from "@vicinae/api";
 import { cleanText } from "./api";
 import { copyToClipboard, relativeTime } from "./utils";
 import type {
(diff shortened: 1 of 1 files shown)


aurelleb:
@knoopx they map to these keybind settings
image
Unless you meant making the extension keybinds customizable.

aurelleb:
overall this is a very good initiative, we should try to have this in most extensions! Thanks @idlip !

PR #161, opened 2026-02-16
feat(npm-manager): adds extension that lets users use vicinae to manage project packages

FredrikMWold:
This extension integrates Vicinae into normal package-management workflows for JavaScript and TypeScript projects.

It lets you install, uninstall, and update dependencies from Vicinae while still using the project’s existing package manager (npm, pnpm, or bun).

### How it works
The extension is launched with the current project path (pwd) from terminal context. From there, it reads package.json detects the package manager via lockfiles, and runs commands in that project directory.

It provides three commands: npm-install, npm-uninstall, and npm-update.
Install supports npm registry search and selection, uninstall/update work from current dependencies.


FredrikMWold:
Looks like i missed some `"`, but I agree will change it to shell functions so that I don't need do to so much escaping.

If we rename the extension from npm-manager to just npm then i think adding a simple npm search command would make sense.

aurelleb:
In this case, maybe you should rename the project commands so that it's clear that this is to manage the current project, because it may not be clear when you install the extension. I know what they do because I discussed it with you, but it's not clear for most users what this actually does, I think. and you could also think for later about a command that would manage your global packages that could also be useful

FredrikMWold:
The first iteration worked more like that, but was changed for a few reasons.

I want the commands to reflect normal terminal usage:
- `npm install` → `npmi`
- `npm uninstall` → `npmr`
- `npm update` → `npmu`

Both install and uninstall actions were unreliable when triggered immediately for each package. It worked much better to let users first select the packages with the primary action, then perform the install or uninstall in a single, batched secondary action.

The actions became a bit confusing when some selected packages were already installed while others were not. I found it easier to use when the decision to install or uninstall was made first, at the beginning.

Personally, I rarely install and uninstall packages at the same time it’s usually one or the other.

PR #174, opened 2026-03-04
feat(omarchy): parity with latest version

Mrid22:
updated the omarchy extension to add all new options ( except for the visual bg selector )


Mrid22:
finally fixed all the typos, looks good now i think

PR #196, opened 2026-03-15
Aww-switcher/Better random selection and faster loading

C4n4m1:
About the faster loading in grid view :
I'm not totally satisfied , I feels like what I want ideally would need to modify how vicinae renders picture in grid view.
Wwith these changes:  instead of fetching wallpapers and their metadata every time the command is called ( 200 ms min for 5 pictures to 1.4 sec max for ~30 pictures ) 
we do it once and then store it in local storage. Now we only check the hash of the source directory ( ~ 2ms to ~ 20 ms ),
and fetch all infos if the hash has changed .

### DIFF
--- a/extensions/awww-switcher/src/wpgrid.tsx
+++ b/extensions/awww-switcher/src/wpgrid.tsx
@@ -5,7 +5,7 @@ import { WindowManagement as wm } from "@vicinae/api";
 import { omniCommand } from "./utils/hyprland";
 import { createHash } from "node:crypto";
 import { Metadata } from "@vicinae/api/dist/api/components/metadata";
-
+import untildify from "untildify";
 // test
 
 export default function DisplayGrid() {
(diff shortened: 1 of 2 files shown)


0x4c756e61 (approved):
Everything looks alright to me. You're hashing the paths only tho, which could result in the view not getting updated if one of the images are replaced. In practice, I don't think this will cause any problem, but I prefer noting it down

0x4c756e61:
Not specifically related to this pull request, but I'd still like to comment about it. It seems like the slowest part of the extension is actually opening the images and reading their metadata, while here that is bypassed by your caching system on subsequent runs, the first one is still going to become quite slow pretty fast. So there's still room for improvement.

Possible solutions would include rendering the wallpapers at the same time as we fetch their metadata and update the view once we have the metadata, this would result in the user getting near-instant feedback on the very first run.

0x4c756e61:
> * I will see how  adding  last modification date to the hashing affects speed.
> 
> * I think I understand the approach : We should send the list with empty metadata and with a second function in useEffect populate the metadata later.

Yeah I think that could do it

PR #210, opened 2026-03-23
feat: show all icons alphabetically on empty search

nanoraptor:
- Modified filterIconIndex to return all icons sorted alphabetically when search is empty
   - Updated display logic to show all icons instead of 'Start searching' message
   - Updated test to verify new behavior
   - All 11 tests passing


aurelleb:
there are conflicts atm. besides, is this ready or not?

PR #215, opened 2026-03-24
(feat): Add "Gnome extensions" extension

7system7:
**The original idea was to replace [this GNOME extension](https://github.com/G-dH/extensions-search-provider).**

This is a quiet new idea to get information about installed GNOME extensions. It has some other functions to handle them

1. enable/disable them
2. check their screenshot
3. check their all important information 
4. Open their preferences window.
etc.

Maybe, later, I would implement the installation/uninstallation feature from the official repo.

gnome-extensions-1

gnome-extensions-2

gnome-extensions-3


7system7:
Is it OK, or do you need something else? :smiley:

extensions/gnome-extensions/src/gnome-extensions.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  return (
+    <List
+      isLoading={isLoading}
+      isShowingDetail={isShowingDetail}
+      searchBarPlaceholder="Search extensions..."
+      actions={(
+        <ActionPanel>
+          <Action title="Reload Extensions" icon={Icon.RotateAntiClockwise} onAction={loadExtensions} />
+          <Action title="Open Extension Manager" icon={Icon.AppWindow} onAction={() => open("gnome-extensions")} />
```
aurelleb: I'm pretty sure this won't work, as `open` is meant to work with a URL or an app.

extensions/gnome-extensions/package.json
```diff
@@ -0,0 +1,60 @@
...
+      "name": "gnome-extensions",
+      "title": "GNOME Extensions",
+      "subtitle": "Manage GNOME Shell extensions",
+      "description": "List installed GNOME extensions, enable/disable them, open preferences dialog etc.",
+      "mode": "view"
+    }
+  ],
+  "preferences": [
+    {
+      "name": "extensionManagerPath",
```
aurelleb: this doesn't seem to be used anywhere in the extension

aurelleb (changes requested):
these things aside, everything looks good to me! Sorry for the late review.

extensions/gnome-extensions/src/gnome-extensions.tsx
7system7: You are right. I deleted them.

extensions/gnome-extensions/package.json
7system7: I deleted this too. I am going to commit these in a couple of minutes.

7system7:
Sorry for the late answer. There was busy weeks...

aurelleb:
thanks!

PR #224, opened 2026-03-27
feat(iconify-extension): add iconify extension

nino-mau:
This is a port of this raycast extension https://www.raycast.com/destiner/iconify 

It basically allows you to search and browse icons from [Iconify Design](https://iconify.design/) using their api.

### Commands

- **Search All Icons**, used to search all icons, you can filter by specific set using this query format `<set-name>:<icon-name>`

- **View Icons**, used to browse icons by icon set


aurelleb:
thanks!

PR #228, opened 2026-03-28
feat(speedtest): add Ookla Speedtest extension

Oztturk:
## Summary

- Adds a new **Speedtest** extension using the official [Ookla Speedtest CLI](https://www.speedtest.net/apps/cli) (`ookla-speedtest-bin`)
- Real-time animated bar chart (SVG) showing live download and upload speeds as the test runs
- Uses `--format=jsonl` streaming to update the UI on every bandwidth sample (~100ms intervals)
- Blue bar for download, fuchsia/pink-purple bar for upload
- Metadata panel shows ping, jitter, packet loss, latency IQM, ISP and server info
- Share URL opens the result on speedtest.net

## Requirements

The user must have the Ookla Speedtest CLI installed:
```
paru -S ookla-speedtest-bin
```

The extension handles the missing CLI gracefully with an error message and install instructions.

## Test plan

- [ ] Run the extension — bar chart animates in real-time during download and upload phases
- [ ] Final results show correct Mbps values (verified against speedtest.net website)
- [ ] Ping / jitter / packet loss shown correctly in metadata panel
- [ ] "Open Result in Browser" action opens the share URL
- [ ] Copy download/upload speed actions work
- [ ] Run Again (⌘R) reruns the test
- [ ] Missing CLI shows helpful error message

🤖 Generated with [Claude Code](https://claude.com/claude-code)


extensions/speedtest/src/run-speedtest.tsx
```diff
@@ -0,0 +1,459 @@
...
+  const runTest = useCallback(() => {
+    procRef.current?.kill();
+
+    let cur: SpeedState = { ...INITIAL };
+    setState(cur);
+
+    let lineBuffer = "";
+    let stderrBuffer = "";
+
+    const proc = spawn("speedtest", ["--format=jsonl", "--accept-license", "--accept-gdpr"]);
```
aurelleb: you need to check whether the executable exists before this, and display an appropriate error message. Do not check the output of the `spawn` call to figure out whether the executable was found or not, that's very unreliable.

aurelleb (changes requested):
it doesn't work for me, it's just stuck on this view, with no error message in the UI or in the console:
image

extensions/speedtest/src/run-speedtest.tsx
Oztturk: My bad i fixed the issue and pushed a new commit. Thanks for the correction. Please let me know if i missed anything else

PR #230, opened 2026-03-29
feat: add Ollama Translate extension

AdityaZxxx:
## Summary

Translate text using local language models with Ollama directly from Vicinae. A privacy-focused alternative to Google Translate

## Features
- **82 languages** supported
- **3 Commands:** translate, instant-copy, instant-paste
- **Auto-detect source language**
- **Multiple Ollama models support**

## Requirements
- Ollama installed locally
- At least one Ollama model downloaded

## Testing
- Build: ✅ Pass
- Lint: ✅ Pass


aurelleb:
really cool! I think you should probably circumvent the preference system and present the user with a view where they can pick the model they want (and suggest gemma:1b). Because otherwise, the default experience may be very bad if it picks any other model (assuming gemma:1b is not available)

extensions/ollama-translate/package.json
```diff
@@ -0,0 +1,85 @@
...
+  "keywords": [
+    "translation",
+    "ollama",
+    "language",
+    "ai",
+    "local",
+    "privacy"
+  ],
+  "license": "MIT",
+  "author": "AdityaRahmad",
```
aurelleb, line 19: author name needs to match your Github account's username

aurelleb (changes requested):

PR #235, opened 2026-04-01
feat: Add Noctalia-Shell Walllpaper Selector Extension

Gimblet:
# Noctalia-Shell Wallpaper Selector

This simple extension helps user's quickly view and change their wallpaper if they use Noctalia-Shell as their wallpaper manager/provider. There is a DEMO and additional information in the README.

This extension pretty much only uses the Vicinae API alongside a shell executor from Node JS for executing the wallpaper change via the Noctalia-Shell IPC. 

The extension currently exposes only one command but more are planned on the future using more Noctalia-Shell IPC functions. 

The extension requires two preferences to be set by the user:

- The absolute path where the user stores wallpapers. (Example: /home/gimblet/Wallpapers)
- The name of their display where the wallpaper will change (Example: DP-1).

### Requirements

- [Noctalia-Shell](https://github.com/noctalia-dev/noctalia-shell)
- Vicinae 

✔️ This extension passes the lint check


aurelleb:
thx!

PR #242, opened 2026-04-10
fix: accept all inputs in outputs function

amadejkastelic:
Fix regression caused by #232, by accepting all arguments in outputs.

```
       error: function 'outputs' called with unexpected argument 'flake-compat'
       at /nix/store/p427p1rnbwdj1mj2k9axc2x9hfkmh54c-source/flake.nix:29:5:
           28|   outputs =
           29|     {
             |     ^
           30|       self,
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,7 @@
       nixpkgs,
       systems,
       vicinae,
+      ...
     }:
     let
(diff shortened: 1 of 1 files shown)


aurelleb:
thx, thanks for the quick fix

PR #243, opened 2026-04-14
Added per setting icons for GNOME Settings Plugin as well as a new Icon

shahiljain:
I thought since the settings are now shown in universal search it would be easier to identify with their gnome specific icons. Also made the extension icon a gnome settings type icon. I have also coloured each of the settings differently based on which section they belong to.


aurelleb:
I really like this change, thanks. better than a single unified icon.

PR #251, opened 2026-04-22
fix(pulseaudio): optional application.name

veryard:
Was encountering issue with Spotify not having an application.name and would cause the extension to throw "Failed to read PulseAudio State". Doing this resolved the issue for me.


aurelleb:
thanks!

PR #272, opened 2026-05-14
vscode-recents: Fix DB paths

### LINKED ISSUE #264: [vscode-recents] VS Code Recents empty on recent VS Code versions
### Extension name

vscode-recents

### Description

The VS Code Recents extension is currently showing an empty list for me, although VS Code itself still shows recent projects correctly.

I am not completely sure this is the root cause, but from local debugging it looks like this may be related to recent VS Code storage changes.

What I checked:

- Vicinae logs only show:

```text
[EXTENSION] "Worker exited with code 1"
````

- Querying the old VS Code state DB returns no row:

```fish
sqlite3 ~/.config/Code/User/globalStorage/state.vscdb \
    "select value from ItemTable where key = 'history.recentlyOpenedPathsList';"
```

- Querying the new shared storage DB does return my recent projects:

```fish
sqlite3 ~/.vscode-shared/sharedStorage/state.vscdb \
    "select value from ItemTable where key = 'history.recentlyOpenedPathsList';" | jq .
```

This seems similar to issues/fixes in other projects after VS Code 1.118 moved recently opened history to shared application storage:

- VS Code change: https://github.com/microsoft/vscode/pull/311963
- Similar Raycast extension issue: https://github.com/raycast/extensions/issues/27537
- Similar PowerToys Run VS Code Workspaces fix: https://github.com/microsoft/PowerToys/pull/47505

Possible fix, if this diagnosis is correct: for recent VS Code versions, read from `~/.vscode-shared/sharedStorage/state.vscdb` first, then fall back to the legacy `~/.config/Code/User/globalStorage/state.vscdb`.

### Steps to reproduce

1. Use VS Code 1.118+ with some recent projects visible in VS Code.
2. Install/open the Vicinae VS Code Recents extension.
3. Open the Recents command/view.

## Expected behavior

The extension lists the same recent projects that VS Code shows.

## Actual behavior

The Recents view is empty.

### Vicinae version

v0.20.15 (commit a0b875caf)

ShyAssassin:
Fixes #264
Microsoft sucks and blatantly disregards XDG standards they *used* to follow.

Pinging @PiasekDev for testing

### DIFF
--- a/extensions/vscode-recents/package-lock.json
+++ b/extensions/vscode-recents/package-lock.json
@@ -914,7 +914,6 @@
       "integrity": "sha512-bbPBYYrtZbkt6Os6FiTLCTFxvq4tt3JKall1vRwshA3fdVztsLAatFaZobhkBC8/BrPetoa0oksYoKXoG4ryJg==",
       "hasInstallScript": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
@@ -3414,7 +3413,6 @@
       "version": "4.0.2",
       "inBundle": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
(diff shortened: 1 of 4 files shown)


PiasekDev:
Tested locally on Vicinae v0.21.0 with VS Code 1.120.0 on Arch Linux and it seems to be working as expected: the recents shown by the extension match VS Code’s own recents list from what I can tell.

Thanks for the fix!

aurelleb:
thx!

PR #273, opened 2026-05-17
feat: VScrot - multi-platform screenshot extension for Vicinae

JaINTP:
## feat: VScrot - multi-platform screenshot extension for Vicinae

VScrot is a new Vicinae extension for capturing, annotating, and managing screenshots across Linux (Wayland & X11), macOS, and Windows.

---

### Overview

VScrot provides a Raycast-style extension interface with full tool auto-detection, per-session tool overrides, and a rich `manage-tools` command for discovering and installing missing dependencies.

---

### Commands

| Command | Description |
|---|---|
| `scrot` | Take a screenshot with mode selection (area, window, monitor, all monitors). Preview with save/copy/annotate/discard actions. |
| `area` | Shortcut to capture area mode |
| `window` | Shortcut to capture window mode |
| `monitor` | Shortcut to capture single monitor |
| `all` | Shortcut to capture all monitors |
| `manage-tools` | Browse all supported tools by category with install status and inline install commands |

---

### Capture Backends

Auto-detected in priority order, overridable via preferences or inline tool picker:

- **Wayland:** `grimblast`, `grim`+slurp, `spectacle`, `gnome-screenshot`, `flameshot`
- **X11:** `maim`+slop+xdotool, `scrot`
- **macOS:** `screencapture` (built-in), `screenshot-desktop`
- **Windows:** `screenshot-desktop`, `flameshot`

---

### Annotation Tools

Post-capture annotation with two modes:
- **Auto-reload** (`satty`, `swappy`): blocks until the tool closes, then refreshes the preview automatically
- **Manual save** (`GIMP`, `Pinta`, `mspaint`): opens in background; user saves, then taps **Refresh Preview**

---

### Clipboard Support

Platform-native clipboard with automatic fallback: `wl-copy` → `xclip` → `pbcopy` → PowerShell

---

### Files Added

- `src/scrot.tsx`: main command (capture mode list + preview)
- `src/area|window|monitor|all.tsx`: shortcut entry points
- `src/manage-tools.tsx`: tool discovery and install helper
- `src/backends/`: 10 capture backend implementations
- `src/annotators/`: 6 annotation tool implementations
- `src/lib/`: capture, annotate, clipboard, filesystem, preferences, and tool-selection utilities
- `src/components/`: `CaptureList`, `PreviewDetail`, `ToolPicker` UI components
- `src/lib/tools-database.ts`: full tool registry with install commands for pacman, apt, dnf, brew, winget
- `README.md`: full user-facing documentation

---

### Stats

`43 files changed, 6416 insertions(+)`


aurelleb:
@clankus-aurelius review

clankus-aurelius:

aurelleb:
@clankus-aurelius review

extensions/vscrot/src/lib/filesystem.ts
```diff
@@ -0,0 +1,57 @@
+import fs from "node:fs";
+import path from "node:path";
+import os from "node:os";
+import { showToast } from "@vicinae/api";
+import type { Preferences } from "./preferences";
+import { expandPath } from "./preferences";
+import { formatDateTokens } from "./dateFormat";
+
+export const TEMP_PATH = path.join(os.tmpdir(), "vscrot_last.png");
```
clankus-aurelius, line 9: **🔴 Blocking — Screenshots use a predictable shared temporary path**

Rule: `SECURITY-003`

Every capture is written to `/tmp/vscrot_last.png`, where it remains after saving or discarding. On multi-user systems this predictable ad-hoc file can expose sensitive screenshot contents and permits races around its replacement.

**Suggested resolution:** Create a unique, permission-restricted temporary directory or file for each capture and remove it after saving, discarding, cancellation, and command teardown.

extensions/vscrot/src/scrot.tsx
```diff
@@ -0,0 +1,186 @@
...
+		await saveAnnotatorChoice(id);
+		setActiveAnnotatorId(id);
+	};
+
+	const handleDeleteFile = (filePath: string) => {
+		fs.unlinkSync(filePath);
+		refreshRecent();
+	};
+
+	if (lastCapture) {
```
clankus-aurelius, line 146: **🔴 Blocking — Permanent deletion duplicates the Vicinae trash API**

Rule: `API-001`

Recent screenshots are deleted directly with `fs.unlinkSync`. The current API declares `trash(path): Promise<void>`, which completely covers user-facing file removal while remaining portable and recoverable.

**Suggested resolution:** Import `trash` from `@vicinae/api`, await `trash(filePath)`, handle failures, and then refresh the list.

extensions/vscrot/src/backends/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+/** Escape a path for safe use inside double-quoted shell strings. */
+export const shellEscape = (p: string): string => p.replace(/"/g, '\\"');
+
+export const isCommandAvailable = (cmd: string): boolean => {
+	try {
+		execSync(`command -v ${cmd}`, { stdio: "ignore" });
+		return true;
+	} catch {
+		return false;
+	}
```
clankus-aurelius, line 12: **🔴 Blocking — Tool detection cannot work on Windows**

Rule: `CORRECTNESS-001`

`isCommandAvailable` always executes the POSIX shell builtin `command -v`. On Windows this reports tools such as Flameshot, GIMP, and Pinta as unavailable, breaking the advertised Windows area capture and annotation workflows.

**Suggested resolution:** Use platform-appropriate executable discovery, such as `where.exe` on Windows and `command -v` on POSIX systems, with parameterized process arguments.

extensions/vscrot/src/backends/screenshot-desktop.ts
```diff
@@ -0,0 +1,26 @@
...
+
+	capture: async (mode: CaptureMode, outputPath: string) => {
+		if (mode === "monitor") {
+			const displays = await screenshot.listDisplays();
+			await screenshot({
+				screen: displays[0]?.id,
+				filename: outputPath,
+				format: "png",
+			});
+		} else {
```
clankus-aurelius, line 22: **🔴 Blocking — Monitor capture always selects the first display**

Rule: `CORRECTNESS-001`

For every `monitor` request, the backend passes `displays[0]?.id`; no user-selected display reaches this code. The advertised “Capture a selected monitor” command therefore captures the first display on Windows and macOS.

**Suggested resolution:** Present the displays returned by `listDisplays`, retain the selected display ID, and pass that ID to the backend capture call.

extensions/vscrot/src/area.tsx
```diff
@@ -0,0 +1,87 @@
...
+		null,
+	);
+
+	const refreshPreview = () => {
+		setLastCapture(null);
+		setTimeout(() => setLastCapture(TEMP_PATH), 100);
+	};
+
+	const capture = async (backendId: string) => {
+		const result = await captureScreenshot("area", backendId);
```
clankus-aurelius, line 27: **🔴 Blocking — Shortcut commands use backends that do not support their mode**

Rule: `CORRECTNESS-001`

The shortcut blindly uses the globally resolved backend. For example, a default Windows installation resolves bundled `screenshot-desktop`, which supports only full and monitor capture, so the advertised Capture Area command always fails; analogous incompatibilities affect other shortcuts.

**Suggested resolution:** Resolve an installed backend that supports the shortcut's requested mode, or show an actionable unavailable-state view when none exists.

extensions/vscrot/src/backends/index.ts
```diff
@@ -0,0 +1,32 @@
...
+	grimblastBackend,
+	grimBackend,
+	spectacleBackend,
+	gnomeScreenshotBackend,
+	flameshotBackend,
+	maimBackend,
+	x11ScrotBackend,
+	screencaptureBackend,
+	screenshotDesktopBackend,
+];
```
clankus-aurelius, line 25: **🟠 Warning — Auto-detection ignores the active display environment**

Rule: `CORRECTNESS-001`

Backends are selected solely by executable availability and fixed priority. An installed Wayland tool can therefore be selected during an X11 session, or an X11 tool during Wayland, causing capture to fail despite another compatible backend being installed.

**Suggested resolution:** Include platform and session compatibility in backend availability checks before applying the priority order.

extensions/vscrot/src/lib/capture.ts
```diff
@@ -0,0 +1,42 @@
...
+			await new Promise((resolve) => setTimeout(resolve, delay * 1000));
+		}
+		await backend.capture(mode, TEMP_PATH, outputName);
+		exec("vicinae open");
+		return TEMP_PATH;
+	} catch {
+		// User cancelled selection (e.g. ESC in slurp) or tool error - surface nothing
+		exec("vicinae open");
+		return null;
+	}
```
clankus-aurelius, line 41: **🟠 Warning — Capture tool failures are silently treated as cancellation**

Rule: `UX-001`

All backend errors—including missing helper programs and execution failures—are swallowed without feedback. Users are returned to Vicinae with no explanation or recovery action.

**Suggested resolution:** Distinguish known cancellation outcomes from operational errors and show an actionable HUD or toast for the latter.

extensions/vscrot/src/lib/annotate.ts
```diff
@@ -0,0 +1,43 @@
...
+		annotator.annotate(imagePath);
+	} catch (e) {
+		console.error("Failed to launch annotator", e);
+		showHUD(`Failed to open ${annotator.displayName}`);
+		return false;
+	}
+	showToast({
+		title: `Opened in ${annotator.displayName}`,
+		message: "Save the file there, then use Refresh Preview to reload.",
+	});
```
clankus-aurelius, line 41: **🟠 Warning — Manual annotator launch failures report success**

Rule: `UX-001`

The annotator promise is not awaited, so its rejection cannot reach this `try` block. The code consequently displays “Opened in …” even when the editor failed to launch; the annotator implementations also resolve regardless of `exec` errors.

**Suggested resolution:** Await `annotator.annotate(imagePath)` and make each manual annotator reject when its child process reports a launch error before showing success.

extensions/vscrot/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Clipboard Tools
+
+Images are copied to the clipboard using platform-native tools with automatic fallback:
+
+| Tool | Platform | Link |
+|---|---|---|
+| **wl-copy** (wl-clipboard) | Linux Wayland (primary) | [wl-clipboard](https://github.com/bugaevc/wl-clipboard) |
+| **xclip** | Linux X11 (fallback) | [xclip](https://github.com/astrand/xclip) |
+| **pbcopy** | macOS (built-in) | - |
+| PowerShell clipboard | Windows (built-in) | - |
```
clankus-aurelius, line 103: **🟠 Warning — Clipboard documentation describes dependencies the implementation does not use**

Rule: `MANIFEST-001`

The README claims native-tool fallback through `wl-copy`, `xclip`, `pbcopy`, and PowerShell, while the implementation exclusively uses Vicinae's declared `Clipboard.copy` API. The tool manager consequently also encourages installation of unnecessary clipboard utilities.

**Suggested resolution:** Document Vicinae clipboard handling and remove the unused clipboard tools from the installation database and bulk install script.

```suggestion
## Clipboard Support

Images are copied through Vicinae's `Clipboard.copy` API; no separate clipboard utility is required.
```

extensions/vscrot/src/monitor.tsx
```diff
@@ -0,0 +1,152 @@
...
+				onCopy={() => copyToClipboard(lastCapture, prefs.autoclose_vicinae)}
+				onAnnotate={handleAnnotate}
+				onRefreshPreview={refreshPreview}
+				onReshoot={() => {
+					setLastCapture(null);
+					setPhase("selecting");
+				}}
+				onDiscard={() => setLastCapture(null)}
+			/>
+		);
```
clankus-aurelius, line 125: **🟠 Warning — Discard leaves monitor capture on a blank screen**

Rule: `CORRECTNESS-001`

Discard clears `lastCapture` but leaves `phase` set to `preview`. Neither rendering branch then matches, so the command returns `null` permanently.

**Suggested resolution:** When discarding, also return the phase to `selecting` or close the command.

```suggestion
onDiscard={() => {
	setLastCapture(null);
	setPhase("selecting");
}}
```

clankus-aurelius (changes requested):
The extension has publication blockers around screenshot privacy, broken cross-platform capture commands, and duplicated native file deletion. Several failure and state-management paths also need correction.

---
_Automated review found 5 publication-blocking issues._

extensions/vscrot/src/lib/capture.ts
```diff
@@ -0,0 +1,60 @@
...
+		showHUD(`${backend.displayName} does not support "${mode}" capture`);
+		return null;
+	}
+
+	// Native hand-off writes straight to the save directory: the desktop tool's
+	// notification then refers to a file that persists, instead of a temporary
+	// one the extension deletes.
+	const native = isNativeHandoff();
+	const target = native ? getSavePath(getPrefs()) : TEMP_PATH;
+	if (native) mkdirSync(dirname(target), { recursive: true });
```
clankus-aurelius, line 37: **🔴 Blocking — Native hand-off breaks the main command's default workflow**

Rule: `CORRECTNESS-001`

Native mode returns the final save path, but `scrot.tsx` still runs Vicinae annotation, copy, and save processing. With the default `save_to_file` enabled, it calls `copyFileSync` with identical source and destination paths and can crash after capture; otherwise it still shows the Vicinae preview despite the preference promising no preview.

**Suggested resolution:** Return capture ownership information and make the main command skip annotation, copy, save, and preview processing when native hand-off owns the result.

extensions/vscrot/src/backends/index.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Which display environments each backend can actually drive. Availability on
+ * PATH is not enough: an installed Wayland tool cannot capture an X11 session
+ * and vice versa, so a compatibility check has to run before the priority order.
+ */
+const BACKEND_SESSIONS: Record<string, Session[]> = {
+	grimblast: ["wayland"],
+	grim: ["wayland"],
+	spectacle: ["wayland", "x11"],
+	"gnome-screenshot": ["wayland", "x11"],
```
clankus-aurelius: **🔴 Blocking — Windows Flameshot is excluded from backend selection**

Rule: `CORRECTNESS-001`

Flameshot is declared compatible only with Wayland and X11, although the extension advertises it as the Windows area-capture backend. On Windows, mode-aware resolution rejects Flameshot and no bundled backend supports area capture.

**Suggested resolution:** Include `win32` in Flameshot's compatible sessions and ensure its Windows capture invocation is supported.

```suggestion
	flameshot: ["wayland", "x11", "win32"],
```

extensions/vscrot/src/backends/index.ts
```diff
@@ -0,0 +1,139 @@
...
+	flameshot: ["wayland", "x11"],
+	maim: ["x11"],
```
clankus-aurelius: **🔴 Blocking — The scrot backend ID never matches its session entry**

Rule: `CORRECTNESS-001`

The backend ID is `scrot`, but the compatibility table uses `x11-scrot`. Consequently scrot is always rejected by mode-aware selection, leaving X11 users with only scrot installed unable to capture.

**Suggested resolution:** Key the compatibility entry by the backend's actual `scrot` ID.

```suggestion
	scrot: ["x11"],
```

extensions/vscrot/src/backends/spectacle.ts
```diff
@@ -0,0 +1,79 @@
...
+			stableFor = 0;
+		}
+		lastSize = size;
+		await sleep(POLL_MS);
+	}
+
+	// No file within the timeout: the user cancelled, or Spectacle failed after
+	// the process we spawned had already exited successfully.
+	throw new Error("Spectacle produced no screenshot");
+};
```
clankus-aurelius: **🟠 Warning — Cancelling Spectacle leaves Vicinae closed for two minutes**

Rule: `CORRECTNESS-001`

The file polling loop waits the full 120-second timeout whenever an interactive Spectacle capture is cancelled, because cancellation produces no output file and process exit is ignored.

**Suggested resolution:** Detect cancellation or early completion and terminate the wait promptly while retaining a reasonable timeout for an active selection.

extensions/vscrot/src/lib/capture.ts
```diff
@@ -0,0 +1,60 @@
...
+		exec("vicinae open");
+		// A cancellation (ESC in slurp/slop) exits non-zero, writes no file and
+		// says nothing on stderr. Anything else is a real failure worth reporting.
+		const stderr = String(
+			(e as { stderr?: Buffer | string }).stderr ?? "",
+		).trim();
+		if (existsSync(target) || stderr !== "") {
+			const detail = stderr.split("\n").pop() || (e as Error).message;
+			showHUD(`${backend.displayName} failed: ${detail}`);
+		}
```
clankus-aurelius: **🟠 Warning — Operational errors without stderr remain silent**

Rule: `UX-001`

Feedback is shown only when the target exists or the exception has `stderr`. Errors such as the new `Spectacle produced no screenshot` exception have neither, so users receive no actionable failure detail.

**Suggested resolution:** Treat only explicitly recognized cancellation outcomes as silent and show the exception message for every other failure.

extensions/vscrot/README.md
```diff
@@ -0,0 +1,141 @@
...
+| **Paint** (mspaint) | Manual save | Windows (built-in) | - |
+| **None** | - | All | Disables the Annotate action |
+
+Auto-detection priority: satty → swappy → GIMP → Pinta → Paint
+
+---
+
+## Clipboard Tools
+
+Images are copied through Vicinae's own clipboard API (`Clipboard.copy`), so no
```
clankus-aurelius, line 97: **🟠 Warning — Tool manager still recommends unused clipboard utilities**

Rule: `MANIFEST-001`

The corrected documentation says no external clipboard tool is required, but Manage Tools still labels `wl-clipboard` as required, lists `xclip` as a fallback, and includes both in the bulk install script.

**Suggested resolution:** Remove the unused clipboard-tool entries and category from the tool database and generated install script.

extensions/vscrot/src/annotators/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Priority order for auto-detection (auto-reload tools first)
+export const ALL_ANNOTATORS: AnnotatorBackend[] = [
+	sattyAnnotator,
+	swappyAnnotator,
+	spectacleAnnotator,
+	gimpAnnotator,
+	pintaAnnotator,
+	mspaintAnnotator,
+];
```
clankus-aurelius, line 20: **🔵 Suggestion — Documented annotation priority omits Spectacle**

Rule: `MANIFEST-001`

Spectacle is now the third auto-detected annotator, while the README and annotation preference description still state that GIMP follows swappy.

**Suggested resolution:** Add Spectacle to the documented auto-detection order and manual-tool preference description.

clankus-aurelius (changes requested):
Several follow-up changes remain broken across native hand-off, backend selection, cancellation, and tool guidance.

---
_Automated review found 3 publication-blocking issues._

extensions/vscrot/src/backends/spectacle.ts
```diff
@@ -0,0 +1,124 @@
...
+			stableFor = 0;
+		}
+		lastSize = size;
+
+		// Nothing written and nothing left running means the user dismissed the
+		// selector. Waiting out the full timeout there would leave Vicinae closed
+		// for two minutes over a keystroke.
+		if (size < 0 && Date.now() - started > STARTUP_GRACE_MS) {
+			goneFor = spectacleRunning() ? 0 : goneFor + POLL_MS;
+			if (goneFor >= SETTLE_MS) {
```
clankus-aurelius: **🟠 Warning — Running Spectacle instance still defeats cancellation detection**

Rule: `CORRECTNESS-001`

Cancellation is recognized only when `pgrep` finds no Spectacle process. If Spectacle was already open or its D-Bus service remains resident, `goneFor` never advances and cancelling an interactive capture still leaves Vicinae closed until the 120-second timeout.

**Suggested resolution:** Track completion or cancellation of the capture request itself instead of treating every system-wide Spectacle process as evidence that this request remains active.

extensions/vscrot/src/lib/capture.ts
```diff
@@ -0,0 +1,62 @@
...
+		await closeMainWindow();
+		if (delay > 0) {
+			await new Promise((resolve) => setTimeout(resolve, delay * 1000));
+		}
+		await backend.capture(mode, target, outputName, { notify: native });
+		exec("vicinae open");
+		// Some tools exit cleanly when the user dismisses the selector
+		// (gnome-screenshot -a among them) and simply write nothing.
+		return existsSync(target) ? target : null;
+	} catch (e) {
```
clankus-aurelius, line 57: **🟠 Warning — Existing native target is mistaken for a new capture**

Rule: `CORRECTNESS-001`

Native mode does not clear or uniquely reserve `target`, yet success is determined solely with `existsSync`. When the formatted filename already exists and a tool exits successfully without writing after cancellation, the command returns the old screenshot as the new result and reports it saved.

**Suggested resolution:** Determine whether this invocation created or modified the target, or generate a collision-free target before capture; do not use existence alone as the success signal.

clankus-aurelius (approved):
Two correctness issues remain in cancellation handling and native hand-off result detection.

---

clankus-aurelius:
@aurelleb automated review passed for `dfd1574`; this extension is ready for your review.

JaINTP:
Round two is done, and the automated checks passed on dfd1574.

Props where they are due first: most of the heavy lifting between review rounds belongs to @asm0dey. They were building their own screen-grab plugin, spotted this PR, and chose to help get it over the line instead of dropping a competing tool. Their PR (JaINTP/vicinae-extensions#1) knocked out all five round-one blockers and kept right on going.

They actually ran the setup on KDE Wayland (which I had not) and found every command broken there. `grim` got picked automatically just because it was installed, but it speaks wlr-screencopy and KWin does not support that, so it just crashed outright. Spectacle's background mode is D-Bus triggered, meaning the process we spawn gives zero useful feedback: in one run it hung around for 11 seconds after the file was already saved, and in another it bailed before writing it at all. Watching the target file directly slashed capture times from 11.2s down to 0.8s. They also caught that returning `null` from a command renders nothing, so a cancelled grab looked identical to one still running, forever. On top of that, the preview markdown was tacking a cache-buster query onto a local file path, so the renderer searched for `capture.png?1234` and spat out a broken image icon. As far as I can tell, that never worked on any setup.

They also added screen detection for Sway, niri, and wlr-randr (it only supported hyprctl before, leaving the picker completely empty everywhere else) plus the "After Capture" setting. That work is pulled in here pretty much exactly as they wrote it.

The five new commits:

| Commit | Findings |
|---|---|
| `cd573e5` fix: pass paths as argv and report capture failures | capture.ts:57 |
| `c64545b` fix: make scrot and windows flameshot selectable | index.ts:73, index.ts:75 |
| `b48135f` fix: end the spectacle wait when a capture is cancelled | spectacle.ts:45 |
| `7a521e5` fix: skip the vicinae pipeline on native hand-off | capture.ts:36 |
| `dfd1574` docs: drop unused clipboard tools and list spectacle | README.md:96, annotators/index.ts:20 |

One round-two finding had the right conclusion but the wrong mechanism. I'll write that up in its own thread instead of cluttering this space.

There is also a fix nobody asked for. Every backend was building its command as a shell string and leaning on a `shellEscape` helper that only escaped double quotes. Since filenames use custom format strings, a path with `$(...)` or backticks would have executed arbitrary commands instead of writing a file. All capture tools and annotators now pass arguments straight to execve, and `shellEscape` is completely gone. The only shell call left is the fixed `hyprctl | jq | slurp` pipeline, which touches zero user data.

Two warnings from round three are still open, and I'll reply to both. The `capture.ts` one is a real bug on my end, caused by one of the changes above. The Spectacle one is a genuine limitation of the approach I took, and I cannot do much better without a KDE box to test on.

Testing carries the same caveat @asm0dey mentioned: the wlroots path was tested against a live Hyprland session, including named-display capture and screen detection. The Spectacle and Windows flameshot paths compile and make sense on paper, but have not been run live because I do not have KDE or Windows installed here.

extensions/vscrot/src/lib/capture.ts
JaINTP: Resolved in `7a521e5`. The takeaway was spot-on, but the underlying reason was off, so looping in @aurelleb.

The actual defect was that `scrot.tsx` was the lone action missing the `isNativeHandoff()` check present across the four shortcuts. Because of that, native mode kept running annotations, clipboard copies, file saves, and markdown previews on an image the system utility had already handled. It exits early now.

Two specifics from the report do not actually pan out, though.

The claim about calling `copyFileSync` with matching source and target locations is typically incorrect. `getSavePath` generates a fresh filename using `Screenshot_%d-%m-%Y_%H-%M-%S`, which operates on second-level precision, and the user interaction takes time. The system clock has almost certainly ticked forward by the second invocation, resulting in a cloned capture with a newer timestamp rather than a self-referential file overwrite.

The note about potential crashes after grabbing a screenshot is also inaccurate. Running `fs.copyFileSync(x, x)` succeeds without issue and preserves the data, since libuv handles matching source and target paths as a no-op. Verified this directly on Node v26.7.0.

The second point, regarding the preview popping up despite preferences disabling it, was totally accurate. That is the behavior users would have actually caught.

extensions/vscrot/src/lib/capture.ts
JaINTP: Spot on, and this bug is completely on me. It crept in with `cd573e5` when I added a file check to stop a cancelled `gnome-screenshot -a` from counting as a win. In native mode, the target is the actual save destination and nothing wipes or reserves it beforehand, so plain `existsSync` cannot tell a brand-new screenshot from an old one already sitting on disk.

Anyone running a `filename_format` without `%S` would run into this non-stop, and taking two quick shots in the same second triggers it on default settings. Handing back an old image and pretending it saved is way worse than the silent failure I was trying to patch in the first place.

The fix is to check the target's stats before running the capture, then make sure it was actually created or updated after the command ran instead of just checking if the file exists.

extensions/vscrot/src/backends/spectacle.ts
JaINTP: Spot on. The check only bumps `goneFor` when `pgrep` sees zero Spectacle processes running, so an open Spectacle window or an active background D-Bus service sends you right back to waiting out the full two-minute timeout. Since we saw the service hang around for roughly 11 seconds after finishing a write, staying resident is normal behavior rather than a rare edge case, making this fix way more limited than I had hoped.

Grabbing a snapshot of Spectacle PIDs before launching and filtering out existing ones handles an open window, but it falls apart if an active daemon handles the work, since no fresh process ever spawns. Tracking the actual request, like you mentioned, pretty much requires speaking directly to Spectacle over D-Bus.

I do not have a KDE box to test on, so any patch I put together here is flying blind, though the current implementation is still way better than always hitting the maximum timeout. I am down to take a crack at the D-Bus route if we want it landed before release, but I would definitely need someone on KDE to verify it works.

asm0dey: Confirmed on KDE Wayland, and it is worse than an open window: with the D-Bus service resident, `spectacle -r -b` spawns no process at all — the resident instance serves the request, so pgrep can never see the count drop.

Switched to Spectacle's own D-Bus reporting, which `spectacle -b -o` emits too: ScreenshotTaken carries the path written (compared, so a screenshot from Spectacle's own window is not mistaken for ours), ScreenshotFailed carries a real error. Cancel emits nothing, so it is inferred from the name being released. Read via `gdbus monitor --dest org.kde.Spectacle` — no new dependency, file-watch fallback when gdbus is absent.

Live Esc run: CaptureCancelled after 1779ms, instead of the full 120s.

Limitation: if the user already has Spectacle open, the name stays owned and a cancel still falls back to the timeout. No signal does better.

Fix: JaINTP/vicinae-extensions#2 (prepared with Claude Code; output above is from a live Plasma session).

extensions/vscrot/src/lib/capture.ts
asm0dey: Fixed as described — mtime recorded before the capture, compared after.

Worth flagging: the same baseline has to be threaded into Spectacle's file wait, or the fallback sees the pre-existing file, watches its size hold steady, and reports the stale screenshot as fresh. Verified with a stale file planted at the target on both paths.

Fix: JaINTP/vicinae-extensions#2 (prepared with Claude Code).

extensions/vscrot/src/backends/spectacle.ts
```diff
@@ -0,0 +1,218 @@
...
+					const arg = line.match(/'((?:[^'\\]|\\.)*)'/)?.[1];
+					if (!arg || resolvePath(arg) === wanted) finish({ kind: "taken" });
+				} else if (line.includes("ScreenshotFailed")) {
+					const arg = line.match(/'((?:[^'\\]|\\.)*)'/)?.[1];
+					finish({
+						kind: "failed",
+						message: arg ?? "Spectacle reported a failure",
+					});
+				} else if (line.includes("is owned by")) {
+					hadOwner = true;
```
clankus-aurelius, line 126: **🟠 Warning — Resident Spectacle still defeats cancellation detection**

Rule: `CORRECTNESS-001`

Cancellation is inferred only after the D-Bus name was observed as owned and then loses its owner. When Spectacle was already open or its service remains resident, the name never becomes unowned, so cancelling an interactive capture still waits for the 120-second timeout.

**Suggested resolution:** Track cancellation for the specific capture request without relying on the global Spectacle service releasing its D-Bus name, or provide another prompt fallback for the resident-service case.

clankus-aurelius (approved):
The native-target freshness issue is resolved. Spectacle cancellation still times out when its D-Bus service is already resident.

---

PR #274, opened 2026-05-18
feat(vscode-recents): add Windsurf editor support

albertonoys:
Very simple, just added support for another VSCode fork.


albertonoys:
@aurelleb Sorry about that, I was going to fix it but I didn't have the time. Thank you for taking the time to fix the conflict.

PR #275, opened 2026-05-19
feat(agenda): add support for local ics files e.g. vdirsyncer

Keyruu:
The previous version only worked with external ICS files. With this you can also include vdirsyncer files!

### DIFF
--- a/extensions/agenda/src/lib/localPath.ts
+++ b/extensions/agenda/src/lib/localPath.ts
@@ -1,0 +1,14 @@
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+export function isLocalPath(url: string): boolean {
+  return (
+    url.startsWith("/") || url.startsWith("~/") || url.startsWith("file://")
+  );
+}
+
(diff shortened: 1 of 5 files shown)


extensions/agenda/src/lib/localPath.ts
```diff
@@ -0,0 +1,14 @@
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+export function isLocalPath(url: string): boolean {
```
vMohammad24, line 4: why not  !url.startsWith("http")?

Keyruu: so we would assume anything other than a url is a localPath?

vMohammad24: yeah and if it fails to resolve then we throw an error

Keyruu: hmm yeah that could work

vMohammad24 (commented):
lgtm

PR #278, opened 2026-05-23
Add timer extension

HarshNarayanJha:
Adding a simple timer extension. You just need to type the duration and an optional note, and it will ring a bell once the time is over. A list of user defined presets can also be set for quick selection. It shows the running times first, if any, and then presets.

Supports basic natural input parsing for prefixes like `2s`, `5m`, or `10h` and standard formats like `00:23` or `01:56:23`. Actions include `cancel timer`, `copy note` and `cancel all timers`.

The actual timers are done using `systemd-run`, which creates transient time units and run the `notify-send` command to show the notification via Vicinae. I had a simple bash sleep implementation before, which probably has a lot of issues like cpu overhead and missed triggers due to sleep and other factors. systemd already does a good job of managing system timers, ensuring these timers remain robust.

This implementation makes having `systemd` and `libnotify` as hard deps, which are pre-installed on most distros. If needed, I can add a fallback to simple sleep or bash looping implementations.

Command in List
image

<details><summary>Adding a Timer</summary>
<p>
image 
</p>
</details> 

<details><summary>Actions</summary>
<p>
image
</p>
</details>


aurelleb:
very cool one, works exactly as intended! Thanks :)

PR #282, opened 2026-05-29
feat(hypr): add hypr extension

nino-mau:
## Summary
Add a new `hypr` extension for accessing Hyprland information from Vicinae using the `hyprctl` CLI.
This is mostly a port of the read-only functionality from the [niri](https://github.com/vicinaehq/extensions/tree/main/extensions/niri) extension for Hyprland. I also added a keybinds command and application icons in the windows command.
## Commands
- outputs: list active and inactive outputs with focus and copy actions
- monitors: list active monitors with focus and copy actions
- workspaces: list workspaces with focus and copy actions
- windows: list windows with focus and copy actions
- layers: list layer surfaces
- keybinds: list keybindings and their descriptions. currently, `hyprctl -j binds` doesn't return much information and the dispatcher is always `__lua` if you have a lua config.
## Icon
Hypr extension icon


FredrikMWold:
Could it make sense to merge the hyperbinding, hyper monitor and this into one hyprland extension?

nino-mau:
> Could it make sense to merge the hyperbinding, hyper monitor and this into one hyprland extension?

 merging hypr-keybinds makes sense since the `keybinds` command already does the same thing

im not sure about merging `hyprland-monitors`. That extension allow to change hyprland monitor state and can persist those changes to config, that feels beyond the scope of this extension. 
this one is more about inspecting Hyprland state and offering basic actions.

Maybe a better idea is to rename this extension so the scope is clearer ? Something like `hypr-info`  `hypr-tools` or `hypr-inspect` . 

I used `hypr` as the name to match the `niri` extension which is also mostly about inspecting state and offering basic actions

FredrikMWold:
The thought was more that combining means that if you are running hyprland and want an integration with vicinae. Then you would just install the "Hyprland" extenstion instead of multiple seperate ones.

aurelleb:
I agree with @FredrikMWold it would be smart to merge the hyprland monitors functionality into this, so that we can have one big extension for Hyprland adjacent stuff.

Thank you for your contribution :)

PR #288, opened 2026-06-04
add extension: clean-keyboard

xevrion:
Locks your keyboard so you can clean it without triggering keys. Press Ctrl+U to unlock anytime.

Uses EVIOCGRAB to grab all keyboard devices at the kernel level. A small C helper is compiled on first run and cached. Requires the user to be in the `input` group (most distros already do this by default).


extensions/clean-keyboard/src/keyboard-locker.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    release_all();
+    printf("unlocked\\n");
+    fflush(stdout);
+    return 0;
+}
+`;
+
+const CACHE_DIR = path.join(os.homedir(), ".cache", "vicinae-extensions");
```
aurelleb: use supportPath for this instead: https://docs.vicinae.com/extensions/api/environment

aurelleb (changes requested):
very cool one, just one thing to fix :)

extensions/clean-keyboard/src/keyboard-locker.ts
xevrion: done! squashed it back into one commit too

xevrion:
btw the postprocessing CI failure is not from my PR; it's trying to fix `extensions/timer/package-lock.json` and fails on the bot push permissions to my fork, just flagging this so its not confusing. 🙂

aurelleb:
yeah I know I need to fix this, thx

PR #310, opened 2026-06-26
add kaomoji extension

bl4zee1g:
Admittedly this is just quickly vibecoded but I use kaomojis way more often than Emojis so this would be really useful for me, hope to see this get added soon (˶>⩊<˶)


aurelleb:
nice one, thanks for your contribution :saluting_face:

PR #312, opened 2026-07-02
fix: regen locks with peerDeps to fix nix build

Keyruu:
The regen of the locks didn't add peerDeps into the locks, I just regenerated them again now with peerDeps present. Also tried some of the builds and it worked for me again.

### DIFF
--- a/extensions/agenda/package-lock.json
+++ b/extensions/agenda/package-lock.json
@@ -19,6 +19,40 @@
         "vitest": "^4.0.18"
       }
     },
+    "node_modules/@emnapi/core": {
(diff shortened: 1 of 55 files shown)


PatrickDaG:
I think you have to add
```
npmFlags = [ "--legacy-peer-deps" ]
```
to the derivation. Some of the extensions do not build for me without. E.G. the `nix` extension.

aurelleb:
Yeah I'm aware of the `--legacy-peer-deps` thing, it will need a sdk fix later. 
Thanks for the PR

Issue #317, opened 2026-07-06
[zed-recents] Zed Recents SSH not opening correctly


Profiidev:
### Extension name

zed-recents

### Description

When trying to open a recent project via ssh it instead tries to open that path on the local machine. I suspect that this is becuase the path attribute is passed to the zed cli instead of the uri attribute. I also had that same problem when i made a custom patch for the raycast zed extension. https://github.com/vicinaehq/extensions/blob/afb84fe4b5253777ff82db8e19e6cc0c9b7f811f/extensions/zed-recents/src/zed-cli.ts#L42
I can submit a pr and fix the issue

### Steps to reproduce

Try to open a ssh recent project using the extension

### Vicinae version

0.22.0

pavle99:
@Profiidev Sure, if you already experienced this issue and fixed it for Raycast Zed extension you can open a PR. I can work on the fix too soon.

Profiidev:
I created a PR. Actually there was all information already available to build the project uri.

PR #320, opened 2026-07-07
fix(zed-recents): use complete file/ssh uri when opening project

Profiidev:
Build a complete file/ssh uri when opening a project.
This fixes the issue that remote workspaces where opened using local path instead of the remote project #317

### DIFF
--- a/extensions/zed-recents/src/zed-cli.ts
+++ b/extensions/zed-recents/src/zed-cli.ts
@@ -65,7 +65,7 @@ export async function openProject(project: RecentProject): Promise<void> {
 function buildUri(project: RecentProject): string {
     if (!project.remote) {
         const uri = new URL("file:///");
-        uri.pathname = project.path;
+        uri.pathname = project.path.replace(/%/g, "%25");
         return uri.toString();
     }
 
@@ -76,7 +76,7 @@ function buildUri(project: RecentProject): string {
     const isRawIpv6 = project.remote.host.includes(":") && !project.remote.host.startsWith("[");
     const safeHost = isRawIpv6 ? `[${project.remote.host}]` : project.remote.host;
     const uri = new URL(`${project.remote.kind}://${safeHost}`);
-    uri.pathname = project.path;
+    uri.pathname = project.path.replace(/%/g, "%25");
 
     if (project.remote.user) {
         uri.username = project.remote.user;
--- a/extensions/zed-recents/src/zed-cli.ts
+++ b/extensions/zed-recents/src/zed-cli.ts
@@ -64,16 +64,28 @@ export async function openProject(project: RecentProject): Promise<void> {
 
 function buildUri(project: RecentProject): string {
     if (!project.remote) {
-        return `file://${project.path}`;
+        const uri = new URL("file:///");
+        uri.pathname = project.path;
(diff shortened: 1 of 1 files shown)


Profiidev:
@pavle99 can you please review this?

pavle99:
@Profiidev Sure

extensions/zed-recents/src/zed-cli.ts
```diff
@@ -52,6 +62,20 @@ export async function openProject(project: RecentProject): Promise<void> {
     }
 }
 
+function buildUri(project: RecentProject): string {
+    if (!project.remote) {
+        return `file://${project.path}`;
```
pavle99: Converting local paths via raw interpolation can break valid names containing URL delimiters. For eg. `/tmp/client#2` would become path `/tmp/client` plus fragment `#2` (and similarly for ?). Keep local paths unchanged, since Zed accepts regular path arguments, or use `pathToFileURL(project.path).href`.

extensions/zed-recents/src/zed-cli.ts
```diff
@@ -52,6 +62,20 @@ export async function openProject(project: RecentProject): Promise<void> {
...
+    }
+
+    if (project.remote.kind === "ssh") {
+        const userPart = project.remote.user ? `${project.remote.user}@` : "";
+        const portPart = project.remote.port ? `:${project.remote.port}` : "";
+        return `${project.remote.kind}://${userPart}${project.remote.host}${portPart}${project.path}`;
```
pavle99: The SSH URI is assembled without encoding or authority validation. Remote paths containing #/?, reserved characters in usernames, missing hosts, IPv6 hosts, etc... can produce incorrect or invalid targets. We should use URI-aware construction, validate host, and cover absolute paths, ~/ paths, user/port variants, and IPv6.

pavle99 (changes requested):

extensions/zed-recents/src/zed-cli.ts
Profiidev: This works just fine. I tested it using this command: `zeditor file:///home/profidev/Coding/tmp/test#1` and it also works from the extension itself

extensions/zed-recents/src/zed-cli.ts
Profiidev: It is not possible to validate everything. For example we will never know if the target path on the host got deleted without trying to connect. I would assume that most of the data is "valid enough" because it is inserted directly by zed into the database. The reserved chars issue and ipv6 support should work now. ~/ paths should not exist and I can't test it because zed makes all paths absolute in the database

pavle99: Okay, that's fine. Maybe do just one more check, that literal `%` characters are encoded as `%25` for both local and SSH paths and verify a literal `%20` pathname (for eg. `/tmp/client%20docs`). After that, this looks good to approve.

Profiidev: The % thing was still an issue. Should be fixed now. I tested it for all of theses paths remote and local: `test#1  test%  test%20  test%25  testing  'test test'`

pavle99 (approved):

PR #327, opened 2026-07-15
Add YubiKey extension

LLawli:
## YubiKey

Use a YubiKey from Vicinae: **TOTP codes**, **FIDO2 passkeys** and **PIV certificates**.

The distinguishing point is that it is **fully native and dependency-free**. It does
not shell out to `ykman`/`yubikit`, needs no Python, and bundles no native addon. Every
protocol is implemented in pure TypeScript on Node's standard library:

- **OATH and PIV** over CCID, by speaking the `pcscd` `winscard_msg` protocol directly
  over its Unix socket.
- **FIDO2** over HID (`CTAPHID` + canonical CBOR + PIN/UV protocol v1/v2) via
  `/dev/hidraw`.
- The OATH access key is read from the **Secret Service** over a small hand-written
  D-Bus client (imported from `ykman`'s keystore when present, so no password prompt in
  the common case).

The only runtime requirement is a running `pcscd` (present on virtually every Linux
desktop), plus the standard FIDO udev rule for the passkey screen. Missing tools are
detected and reported with install guidance.

### Commands

- **OTP Codes** — lists the TOTP codes with a per-account countdown; the primary action
  pastes the selected code into the focused field. Touch-required accounts ask for the
  touch before pasting.
- **Keys & Certificates** — enumerate/delete FIDO2 passkeys; view and export PIV
  certificates. The PIN is asked in-app and kept only in memory.

### Notes for review

- **Not a duplicate of the existing `otp` extension.** That one is a software OTP
  manager (stores TOTP/HOTP secrets). This one reads codes from a physical YubiKey and
  also handles FIDO2/PIV; it does nothing without the hardware key plugged in.
- **Clipboard history:** because `Clipboard.paste()` has no `concealed` flag (only
  `copy()` does), pasting a code writes it to Vicinae's clipboard history in plaintext.
  The extension removes that entry right after pasting, and degrades gracefully (logs a
  warning, never breaks the paste) if the schema changes. An upstream `concealed` option
  on `paste()` would let us delete this code entirely.
- **Interface in English and Portuguese**, selectable via the `Language` preference.
- Design and protocol details are documented under `extensions/yubikey/docs/`.

### Checks

Passes `npm ci`, `vici build` and `vici lint` locally. Preview images are HTML/CSS
mockups (fictitious data), not screenshots of real accounts.


aurelleb:
@clankus-aurelius review

clankus-aurelius:

extensions/yubikey/src/lib/hid.ts
```diff
@@ -0,0 +1,224 @@
...
+    try {
+      const desc = readFileSync(`/sys/class/hidraw/${name}/device/report_descriptor`);
+      if (desc[0] === 0x06 && desc[1] === 0xd0 && desc[2] === 0xf1) return `/dev/${name}`;
+    } catch {
+      // segue
+    }
+  }
+  return null;
+}
+
```
clankus-aurelius: **🔴 Blocking — Passkey operations can target a non-YubiKey authenticator**

Rule: `CORRECTNESS-001`

When no Yubico device is found, discovery selects the first arbitrary FIDO authenticator. The command can consequently enumerate and permanently delete credentials from a different security key while presenting it as the user's YubiKey.

**Suggested resolution:** Return no device unless the discovered FIDO interface is verified as Yubico, or explicitly expose device selection and identity to the user.

```suggestion
  // No verified Yubico FIDO interface was found.
  return null;
```

extensions/yubikey/src/lib/i18n.ts
```diff
@@ -0,0 +1,371 @@
...
+  "err.pin_blocked": "PIN do PIV bloqueado. É preciso o PUK para destravar.",
+  "err.no_cert": "Não há certificado nesse slot",
+  "err.no_device": "Nenhum autenticador FIDO encontrado",
+  "err.no_access": "Sem permissão para acessar a YubiKey via HID. Instale as regras udev do FIDO (`libfido2` no Fedora/Arch, `libu2f-udev` no Debian/Ubuntu).",
+  "err.ctap.49": "PIN incorreto",
+  "err.ctap.52": "Muitas tentativas seguidas. Retire a YubiKey e insira de novo.",
+  "err.ctap.53": "PIN bloqueado. Retire a YubiKey e insira de novo.",
+  "err.ctap.54": "PIN bloqueado. É preciso redefini-lo (reset do FIDO2).",
+  "err.ctap.unknown": "A YubiKey recusou a requisição (status {status})",
+};
```
clankus-aurelius: **🔴 Blocking — Extension UI includes Portuguese interface text**

Rule: `UX-003`

The Portuguese dictionary supplies non-English titles, actions, messages, errors, and other interface copy. YubiKey management is not inherently language-specific, so selectable non-English UI is not an allowed exception.

**Suggested resolution:** Remove the Portuguese interface option and dictionary, and keep all extension-provided UI copy in English.

extensions/yubikey/src/lib/piv.ts
```diff
@@ -0,0 +1,260 @@
...
+  CERT_INFO: 0x71,
+  METADATA_RETRIES: 0x06,
+} as const;
+
+const PIN_P2 = 0x80;
+const PIN_LEN = 8;
+
+/** Slot → nome legível e id do objeto que guarda o certificado daquele slot. */
+export const SLOTS = [
+  { slot: "9a", name: "AUTHENTICATION", label: "Autenticação", objectId: 0x5fc105 },
```
clankus-aurelius: **🔴 Blocking — PIV slot titles are always Portuguese**

Rule: `UX-003`

These Portuguese labels are rendered directly as List.Item titles even when English is selected.

**Suggested resolution:** Use English labels for the PIV slots.

```suggestion
export const SLOTS = [
  { slot: "9a", name: "AUTHENTICATION", label: "Authentication", objectId: 0x5fc105 },
  { slot: "9c", name: "SIGNATURE", label: "Digital signature", objectId: 0x5fc10a },
  { slot: "9d", name: "KEY_MANAGEMENT", label: "Key management", objectId: 0x5fc10b },
  { slot: "9e", name: "CARD_AUTH", label: "Card authentication", objectId: 0x5fc101 },
] as const;
```

extensions/yubikey/package.json
```diff
@@ -0,0 +1,118 @@
+{
+  "$schema": "https://raw.githubusercontent.com/vicinaehq/vicinae/refs/heads/main/extra/schemas/extension.json",
+  "name": "yubikey",
+  "title": "YubiKey",
```
clankus-aurelius, line 4: **🔴 Blocking — Manifest falsely claims no external dependencies**

Rule: `MANIFEST-001`

The store description says the extension has no external dependencies, while its implementation and documentation require pcscd and require FIDO udev rules for passkey functionality.

**Suggested resolution:** Remove the dependency-free claim or clearly disclose the runtime requirements in the store-facing description.

```suggestion
  "description": "Use a YubiKey for TOTP codes, FIDO2 passkeys, and PIV certificates on Linux",
```

clankus-aurelius (changes requested):
The extension has publication-blocking credential-storage, wrong-device, localization, and manifest issues.

---
_Automated review found 5 publication-blocking issues._

LLawli:
Thanks for the review. All four blocking findings are addressed in `4b2ec53`.

**CORRECTNESS-001 — `src/lib/hid.ts`.** `findFidoDevice()` no longer falls back to an
arbitrary FIDO interface. It now matches the Yubico USB vendor id (`0x1050`) in the sysfs
uevent and returns `null` when no Yubico interface is present, so passkey enumeration and
deletion can never reach another security key. Matching on the vendor id rather than the
device name also avoids trusting a string another vendor controls.

**UX-003 — `src/lib/i18n.ts` and `src/lib/piv.ts`.** The Portuguese dictionary, the
`language` preference and the locale detection are gone; the PIV slot labels use the exact
English wording you suggested. `t()` and `localizeError()` stay as the single place where
user-facing wording lives, now English-only. I also removed `describeCtapError()` and
`describeError()`: both were unused, and they held the last Portuguese strings outside the
dictionary, which `localizeError()` would have shown verbatim through its message fallback.
The source comments are translated too.

**MANIFEST-001 — `package.json`.** The description is the one you suggested. I applied the
same correction to the README, the changelog and `docs/architecture.md`, which repeated the
claim; they now say "no Python and no native addons", which is what it was meant to convey.
The requirements section already lists `pcscd`, `ccid` and the FIDO udev rule.

---

The status comment mentions five blocking findings, but only four inline threads were
posted. The one I cannot see is the credential-storage finding, so I want to flag what the
extension does in case that is the subject and let you tell me if it needs changing.

The only credential the extension persists is the OATH access key, in `src/lib/secrets.ts`.
It goes to the Secret Service over D-Bus, never to disk in plain text. That choice is about
interoperability rather than storage: if the user has already run `ykman oath ... -r`, the
key is in ykman's keystore encrypted with a wrap key that also lives in the Secret Service,
so reading it from there means they never see a password prompt for something they already
authorized. `LocalStorage` could hold the bytes, but it would not reach ykman's existing
entry, and it would move a secret that is currently keyring-encrypted into extension
storage. SECURITY-003 seemed to allow this explicitly, so I left it as is.

The FIDO2 and PIV PINs are never persisted anywhere: they live in the command's memory for
as long as the command is open, and are never passed through argv.

Happy to change any of it if the missing finding says otherwise, and to see its text if it
comes back on the next pass.

extensions/yubikey/src/lib/secrets.ts
```diff
@@ -0,0 +1,238 @@
...
+  } finally {
+    session?.bus.close();
+  }
+}
+
+/** Stores the key derived from a password the user typed. */
+export async function saveAccessKey(deviceId: string, key: Buffer): Promise<void> {
+  memoryCache.set(deviceId, key);
+  let session: Session | null = null;
+  try {
```
clankus-aurelius, line 232: **🟠 Warning — Remembered OATH keys are never persisted**

Rule: `CORRECTNESS-001`

`createItem` uses the D-Bus signature `a{sv}(oayays)b`, but the custom D-Bus writer has no `b` boolean handler. Marshalling therefore throws; `saveAccessKey` silently catches it, leaving the key only in memory despite the enabled “Remember” option.

**Suggested resolution:** Implement boolean marshalling for the D-Bus `b` type and propagate persistence failures so the UI does not claim the key was remembered.

extensions/yubikey/src/security-keys.tsx
```diff
@@ -0,0 +1,425 @@
...
+    fidoSupported: null,
+    loading: true,
+    error: null,
+  });
+
+  // The PIN lives only here, in the command's memory. Never in Cache, LocalStorage or argv.
+  const [fidoPin, setFidoPin] = useState<string | null>(null);
+  const [fidoCreds, setFidoCreds] = useState<FidoCred[] | null>(null);
+
+  const load = useCallback(async () => {
```
clankus-aurelius, line 50: **🟠 Warning — PIV failure blocks the independent FIDO2 workflow**

Rule: `CORRECTNESS-001`

The command awaits PIV initialization before checking FIDO2. If `pcscd`, CCID, or the PIV applet is unavailable, execution jumps to the outer catch and users cannot manage passkeys even though FIDO2 uses the independent HID transport.

**Suggested resolution:** Load PIV and FIDO2 independently, retaining successful results and presenting a separate error state for each section.

extensions/yubikey/src/lib/clipboard.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+export async function pasteAndForget(text: string): Promise<void> {
+  await closeMainWindow();
+  await Clipboard.paste(text);
+
+  if (prefs().purgeClipboardHistory !== false) {
+    await purgeFromHistory(text);
+  }
+}
+
```
clankus-aurelius: **🟠 Warning — Clipboard-history purge failures are hidden**

Rule: `UX-001`

`purgeFromHistory` returns `false` when SQLite is unavailable, the schema differs, or deletion fails, but this result is ignored. The OTP remains in plaintext history while the user receives no warning that the enabled protection failed.

**Suggested resolution:** Check the purge result and show actionable user-facing feedback when the history entry could not be removed.

extensions/yubikey/package.json
```diff
@@ -0,0 +1,105 @@
...
+        "yubikey",
+        "oath"
+      ]
+    },
+    {
+      "name": "security-keys",
+      "title": "Keys & Certificates",
+      "subtitle": "YubiKey",
+      "description": "Manage FIDO2 passkeys, PIV slots and certificates on the YubiKey",
+      "mode": "view",
```
clankus-aurelius, line 40: **🟠 Warning — Command description overstates PIV management**

Rule: `MANIFEST-001`

The command says it manages PIV slots and certificates, but the implementation is read-only and only views or exports certificates.

**Suggested resolution:** Describe PIV functionality as viewing and exporting certificates.

```suggestion
      "description": "Manage FIDO2 passkeys and view or export PIV certificates on the YubiKey",
```

clankus-aurelius (approved):
The extension has four actionable issues: one misleading manifest description, two degraded workflows, and broken credential persistence.

---

clankus-aurelius:
@aurelleb automated review passed for `4b2ec53`; this extension is ready for your review.

LLawli:
All four warnings are addressed in `3e45794`. The credential-persistence one was a real bug —
thanks for catching it.

**`src/lib/secrets.ts` — remembered OATH keys were never persisted.** Confirmed exactly as
described: the `b` in `a{sv}(oayays)b` hit the writer's `default:` branch and threw
`marshal: unsupported signature 'b'` on every `CreateItem`, and `saveAccessKey` swallowed it.
D-Bus booleans now marshal as the uint32 0/1 the spec calls for, read back the same way, and
align to 4 bytes inside arrays. `saveAccessKey` returns whether the key reached the keyring,
`OathSession.unlock` propagates it, and the unlock form now says so when the key could only be
kept in memory (no Secret Service running) instead of implying it was remembered. Verified
against a live gnome-keyring: the item is created, found by `SearchItems`, and read back.

**`src/security-keys.tsx` — PIV failure no longer blocks FIDO2.** The two halves load
independently now, PIV keeps its own error row inside its own section, and the full-screen
empty state is reserved for when both are unavailable. Verified with the PIV session stubbed
to throw `no_daemon`: the passkey section stays fully usable.

**`src/lib/clipboard.ts` — purge failures are surfaced.** `pasteAndForget` returns the outcome
and the OTP screen shows a HUD when the code stayed in the history (the main window is already
closed by then, so a toast would have nowhere to land).

One deliberate nuance there: `purgeFromHistory` now distinguishes `failed` from `absent`. An
entry that never appears is usually Vicinae deduplicating a selection identical to the current
one, so there is nothing in the history to warn about, and warning on every such paste would
train the user to ignore the message. Only genuine failures — `node:sqlite` unavailable, a
schema change, a delete that removes nothing — reach the user. Happy to warn on both if you
would rather err the other way.

**`package.json` — command description.** Applied your suggestion; PIV is view/export only in
this release.

The CI checks are sitting at `action_required`, so they need a maintainer to approve the
workflow run for this fork before they report.

extensions/yubikey/package.json
```diff
@@ -0,0 +1,105 @@
...
+      "required": false,
+      "placeholder": "/run/pcscd/pcscd.comm"
+    },
+    {
+      "name": "serial",
+      "title": "YubiKey serial",
+      "description": "Pin the extension to a specific key. Leave empty to use the first one found.",
+      "type": "textfield",
+      "required": false,
+      "placeholder": "25349859"
```
clankus-aurelius, line 68: **🔴 Blocking — Serial preference does not pin operations to that YubiKey**

Rule: `MANIFEST-001`

The preference promises to pin the extension to a specific key, but PC/SC treats it only as a reader-name hint and falls back when unmatched, while HID discovery ignores it entirely. With multiple YubiKeys, OTP/PIV operations can target the wrong key and passkey deletion can affect another key.

**Suggested resolution:** Strictly resolve the configured serial for both CCID and HID and fail when that key is unavailable, or change the preference wording and behavior so it does not claim pinning.

extensions/yubikey/src/lib/touch.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    withOwnConnection(async (conn) => {
+      // `withOwnConnection` closes it; we keep a way to close early when cancelled.
+      onCancel = () => conn.close().catch(() => {});
+      return conn.transaction(async (t) => {
+        const info = await select(t);
+        if (info.challenge) {
+          const key = await loadAccessKey(info.deviceId);
+          if (!key) throw new OathError("locked", "The OATH application is password-protected");
+          await validate(t, key, info.challenge);
```
clankus-aurelius, line 106: **🟠 Warning — Cancelling a touch corrupts the in-flight PC/SC exchange**

Rule: `CORRECTNESS-001`

Cancellation calls `conn.close()` while `transaction()` is awaiting a TRANSMIT response. `close()` sends DISCONNECT on the same socket and installs another `readExactly` waiter, overwriting the blocked operation's waiter and potentially leaving its promise and socket alive indefinitely.

**Suggested resolution:** Add an abort path that immediately destroys the socket and rejects the active waiter; use it for touch cancellation instead of the graceful `close()` sequence.

extensions/yubikey/src/security-keys.tsx
```diff
@@ -0,0 +1,442 @@
...
+      pivError = err instanceof Error ? err : new Error(String(err));
+    }
+
+    let fidoInfo: FidoInfo | null = null;
+    const fidoSupported = fido().available();
+    if (fidoSupported) {
+      fidoInfo = await fido().info().catch(() => null);
+    }
+
+    setState({
```
clankus-aurelius: **🟠 Warning — FIDO initialization errors are discarded**

Rule: `UX-001`

Any `fido().info()` failure is converted to `null`, so permission, I/O, and protocol failures all become the generic “interface not accessible” state and the actual actionable error is lost.

**Suggested resolution:** Preserve the thrown error in state and display `localizeError(error)` in the unavailable item while retaining the retry action.

extensions/yubikey/src/otp-codes.tsx
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  const touchCreds = state.creds.filter((c) => c.touch || c.type === "HOTP");
+  const autoCreds = state.creds.filter((c) => !c.touch && c.type === "TOTP");
+
+  return (
+    <List
+      isLoading={state.loading}
+      searchBarPlaceholder={t("otp.search")}
+      navigationTitle={t("otp.nav")}
```
clankus-aurelius, line 324: **🟠 Warning — A stale stored OATH key leaves no recovery path**

Rule: `CORRECTNESS-001`

Only `locked` opens the password form. If a persisted or imported access key is stale, validation raises `wrong_password`; the command then falls through to the account list without offering a way to enter and save the current password.

**Suggested resolution:** Route both `locked` and `wrong_password` states to `LockedView`, allowing the user to replace the stale key.

```suggestion
if (errCode(state.error) === "locked" || errCode(state.error) === "wrong_password") {
    return <LockedView onUnlocked={onUnlocked} />;
  }
```

extensions/yubikey/src/otp-codes.tsx
```diff
@@ -0,0 +1,517 @@
...
+      // only buy a silent wait of up to 15s before the key even blinks.
+      const busy = Math.ceil(cooldownRemaining() / 1000);
+      if (busy > 0) {
+        await showToast({
+          style: Toast.Style.Failure,
+          title: t("busy.title"),
+          message: t("busy.message", { s: busy }),
+        });
+        return;
+      }
```
clankus-aurelius, line 244: **🟠 Warning — Paste failures are reported in a closed window**

Rule: `UX-001`

`pasteAndForget()` closes the main window before attempting the paste, but this catch reports failure with a toast. Vicinae documents toasts as notifications inside its window; after closing it, the failure is not reliably visible. `showHUD` is the documented feedback surface after closing the launcher.

**Suggested resolution:** Report the failure with `showHUD`, including the account label and localized error.

```suggestion
await showHUD(`${t("otp.paste.failed")} — ${label}: ${localizeError(err)}`);
```

clankus-aurelius (changes requested):
The configured key is not reliably selected, creating a risk of reading from or deleting credentials on another YubiKey. Four additional workflow and error-feedback defects should also be fixed.

---

LLawli:
Addressed in `61a2d67`. The serial finding was right, and worse than the description
suggests — the preference matched nothing at all.

**`package.json` / PC/SC / HID — the `serial` preference now pins.**

I checked what was actually available before choosing an approach:

- The reader name pcscd reports is `Yubico YubiKey OTP+FIDO+CCID 00 00`. The trailing
  digits are reader index and slot, not identity, so `name.includes(serial)` could never
  match and every configured serial fell through to the first card found.
- The USB descriptor carries no serial either: `/sys/.../usb1/1-2/` has `idVendor`,
  `idProduct` and `product`, but no `serial`, on a 5.7.4 with the OTP interface enabled.

So the serial has to come from the key. The management applet (AID `a000000527471117`,
READ CONFIG `00 1D 00 00 00`, DeviceInfo tag `0x02`) reports it, which is what `ykman`
reads too. Every reader holding a card is now asked for its serial, and the connection
only stands on a match; with no match it fails with an actionable message instead of
using a neighbouring key. A reader held exclusively by another program is skipped rather
than aborting the search, and a non-numeric preference value is rejected up front.

Verified against a real key (serial `25349859`): correct serial connects, a different
serial refuses, a non-numeric value refuses, empty keeps the previous behaviour.

FIDO2 gets no equivalent, and I want to be explicit about the limitation rather than
paper over it. CTAP2 exposes an AAGUID, which identifies the model and not the unit, so
there is no way to honour the preference over that transport. Rather than pick one of two
keys and present it as the user's, the passkey path now refuses whenever more than one
YubiKey is connected — verified with the sysfs layer stubbed to show two. Deleting a
passkey is irreversible, so refusing seemed like the only defensible answer. The
preference description now says exactly this.

**`src/lib/touch.ts` — abort path added.** `close()` sends DISCONNECT and waits, which
installs a second read waiter over the one the in-flight TRANSMIT is blocked on, leaving
that promise unsettled. Cancellation now calls a new `abort()` that destroys the socket
and rejects whatever is pending; `close()` stays for the normal end of a connection.

**`src/security-keys.tsx` — FIDO errors preserved.** The state keeps the error and the
unavailable row shows `localizeError(...)`, so a missing udev rule reads as the
install-`libfido2` message rather than a generic "not accessible".

**`src/otp-codes.tsx` — stale key and paste errors.** `wrong_password` now routes to the
unlock form, so a password changed elsewhere can be replaced instead of leaving the
screen stuck on an empty list. Paste failures report through `showHUD` for the reason you
gave.

One note on the previous round, since it is relevant here: the D-Bus boolean fix means
`CreateItem`'s `replace` flag is finally sent, so replacing a stale key overwrites the old
keyring item instead of adding a second one.

extensions/yubikey/src/otp-codes.tsx
```diff
@@ -0,0 +1,517 @@
...
+      <List>
+        <List.EmptyView
+          icon={Icon.XMarkCircle}
+          title={errCode(state.error) === "busy" ? t("device.busy") : t("device.none")}
+          description={state.error ? localizeError(state.error) : undefined}
+          actions={
+            <ActionPanel>
+              <Action title={t("action.retry")} icon={Icon.ArrowClockwise} onAction={load} />
+            </ActionPanel>
+          }
```
clankus-aurelius, line 311: **🟠 Warning — OTP loading errors can be hidden**

Rule: `UX-001`

The failure view is rendered only for selected PcscError codes and only when no cached credentials exist. With cached credentials, or errors such as bad_serial, no_such_serial, no_applet, protocol, or io, the screen instead shows stale/empty rows without explaining the failure.

**Suggested resolution:** Render actionable feedback for every non-lock loading error, regardless of whether cached credentials are available; cached rows may remain visible alongside the error.

extensions/yubikey/src/security-keys.tsx
```diff
@@ -0,0 +1,445 @@
...
+  const refresh = (
+    <Action title={t("action.reload")} icon={Icon.ArrowClockwise} shortcut={{ key: "r", modifiers: ["cmd"] }} onAction={load} />
+  );
+
+  return (
+    <List
+      isLoading={state.loading}
+      searchBarPlaceholder={t("keys.search")}
+      navigationTitle={t("keys.nav")}
+      isShowingDetail
```
clankus-aurelius: **🟠 Warning — Combined failure view hides the FIDO2 error**

Rule: `UX-001`

When both PIV and FIDO2 fail, the full-screen state displays only state.pivError. The user receives no reason or remediation for the simultaneous FIDO2 failure.

**Suggested resolution:** Include both localized failure reasons in the empty state, or render the two independent unavailable sections with their respective retry guidance.

extensions/yubikey/src/lib/pcsc.ts
```diff
@@ -0,0 +1,550 @@
...
+      if (found === wanted) return;
+      await this.detach();
+    }
+
+    // Every candidate was unreachable rather than merely unmatched: report why, since "not
+    // connected" would send the user looking for the wrong problem.
+    if (lastFailure && lastFailure.code === "busy") throw lastFailure;
+
+    throw new PcscError(
+      "no_such_serial",
```
clankus-aurelius, line 382: **🟠 Warning — Failed PC/SC connections leak resources**

Rule: `CORRECTNESS-001`

connect() opens a socket and may establish a context before readersWithCard(), serial validation, or attachment fails, but it does not close the connection on those paths. Each retry can retain another pcscd socket and context.

**Suggested resolution:** Wrap connection setup and reader selection in a try/catch that calls close() before rethrowing, or otherwise make connect() release every acquired resource on failure.

extensions/yubikey/src/lib/fido-session.ts
```diff
@@ -0,0 +1,50 @@
...
+  private async withDevice<T>(fn: (ctap: Ctap2) => Promise<T> | T): Promise<T> {
+    const dev = new HidDevice();
+    await dev.open();
+    try {
+      const ctap = new Ctap2(dev);
+      await ctap.init();
+      return await fn(ctap);
+    } finally {
+      dev.close();
+    }
```
clankus-aurelius, line 23: **🟠 Warning — Failed CTAPHID initialization leaks the device descriptor**

Rule: `CORRECTNESS-001`

dev.open() runs before the try/finally. If opening the hidraw file succeeds but CTAPHID INIT fails, the descriptor is never closed; repeated retries leak descriptors.

**Suggested resolution:** Place dev.open() inside the try/finally so dev.close() runs for partial initialization failures.

```suggestion
const dev = new HidDevice();
try {
  await dev.open();
  const ctap = new Ctap2(dev);
  await ctap.init();
  return await fn(ctap);
} finally {
  dev.close();
}
```

extensions/yubikey/docs/decisions.md
```diff
@@ -0,0 +1,153 @@
...
+opportunistically from `ykman`'s own keystore: read
+`~/.local/share/ykman/oath_keys.json`, fetch `ykman`'s wrap key from the Secret
+Service, and decrypt the Fernet token — all in `node:crypto`.
+
+**Consequences.** A user who already ran `ykman oath ... -r` never sees a password
+prompt. When no Secret Service is available (e.g. a bare sway/Hyprland session), the
+key stays in memory for the session; persisting it in plaintext is offered only
+behind an explicit, off-by-default preference, never silently. Only the derived key
+is ever stored, never the password. See
+[protocols.md](./protocols.md#secret-service).
```
clankus-aurelius, line 64: **🟠 Warning — Documentation describes a nonexistent plaintext-storage preference**

Rule: `MANIFEST-001`

The ADR says plaintext persistence is offered through an off-by-default preference, but the manifest and implementation provide no such preference and retain the key only in memory when Secret Service is unavailable.

**Suggested resolution:** Remove the nonexistent plaintext-persistence claim and describe the implemented memory-only fallback.

```suggestion
**Consequences.** A user who already ran `ykman oath ... -r` never sees a password
prompt. When no Secret Service is available (e.g. a bare sway/Hyprland session), the
key stays in memory for the session. Only the derived key is ever stored, never the
password. See
[protocols.md](./protocols.md#secret-service).
```

clankus-aurelius (approved):
The extension has four operational defects and one documentation mismatch. No clear publication blocker was found.

---

clankus-aurelius:
@aurelleb automated review passed for `61a2d67`; this extension is ready for your review.

LLawli:
All five addressed in `f887287`. Two of them were regressions I introduced in the previous
round, so thanks for catching them before a maintainer had to.

**`src/lib/pcsc.ts` + `src/lib/fido-session.ts` — resources released on failure.** Both
were real. `connect()` now wraps the whole setup and closes the socket and context before
rethrowing, on every path: no reader, no card, a malformed serial, no matching key.
`dev.open()` moved inside the try/finally, since it opens the hidraw descriptor before
running the CTAPHID INIT that can fail. Measured with `/proc/self/fd`: over 1, 10 and 30
consecutive failed connects the descriptor count no longer grows with the attempts.

**`src/otp-codes.tsx` — every load error is visible now.** This one was mine: I added
`bad_serial` and `no_such_serial` last round and never added them to the device-problem
list, so a mistyped serial silently showed stale cached rows. Rather than extend that
list, any error now renders a headline row above the cached accounts with the localized
reason and a retry, and the empty-cache view covers every error instead of a chosen few.
Cached rows stay visible alongside it, as you suggested.

**`src/security-keys.tsx` — both reasons shown.** Also mine, from the same round. I took
the second option you offered and dropped the combined full-screen state entirely: each
section carries its own error and its own retry. A user could otherwise go fix pcscd while
the passkeys were failing on a missing udev rule, which is exactly the wrong place to
send them.

**`docs/decisions.md` — ADR 4 corrected.** The plaintext-persistence preference never
existed. The ADR now describes what is implemented: memory-only for the session when no
Secret Service is available, and the unlock screen says so rather than implying the key
was remembered.

extensions/yubikey/src/lib/touch.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    cancel = () => finish(() => reject(new OathError("cancelled", "Touch cancelled")), true);
+
+    withOwnConnection(async (conn) => {
+      // Cancelling aborts instead of closing: a touch CALCULATE is still in flight and the
+      // graceful DISCONNECT would queue behind it, stranding the pending read.
+      onCancel = () => conn.abort("The touch was cancelled");
+      return conn.transaction(async (t) => {
+        const info = await select(t);
+        if (info.challenge) {
```
clankus-aurelius, line 103: **🟠 Warning — Early cancellation does not stop the touch operation**

Rule: `CORRECTNESS-001`

If cancellation occurs while `conn.connect()` is pending, `onCancel` is still null. After connecting, the callback is assigned and the transaction proceeds despite the already-settled handle, potentially holding the card for the full touch timeout. `waitForCard()` has the same race.

**Suggested resolution:** Track cancellation independently and abort immediately when the connection becomes available, before starting a transaction. Apply the same handling to `waitForCard()`.

extensions/yubikey/src/lib/clipboard.ts
```diff
@@ -0,0 +1,134 @@
...
+          try {
+            rmSync(join(CLIPBOARD_DATA, offerId), { force: true });
+          } catch (err) {
+            console.warn(`[clipboard] could not remove blob ${offerId}: ${String(err)}`);
+          }
+        }
+
+        return Number(result.changes) > 0 ? "removed" : "failed";
+      } catch (err) {
+        db.close();
```
clankus-aurelius, line 98: **🟠 Warning — Clipboard blob deletion failure is reported as purged**

Rule: `SECURITY-003`

A failed `rmSync` is only logged; the function still returns `removed`. The OTP can remain in plaintext under `clipboard-data` while the UI reports a successful purge.

**Suggested resolution:** Track blob-removal failures and return `failed` so `pasteAndForget` surfaces the existing HUD warning.

extensions/yubikey/src/security-keys.tsx
```diff
@@ -0,0 +1,414 @@
...
+    [fidoPin, load],
+  );
+
+  /** Exports a PIV slot's certificate. No PIN needed: the certificate is public. */
+  const exportCert = useCallback(async (slot: SlotInfo) => {
+    try {
+      const pem = await piv().exportCertificate(objectIdForSlot(slot.slot));
+      const path = join(homedir(), `yubikey-piv-${slot.slot}.pem`);
+      writeFileSync(path, pem, { mode: 0o600 });
+      await showToast({ style: Toast.Style.Success, title: t("keys.piv.exported"), message: path });
```
clankus-aurelius: **🟠 Warning — Certificate export silently overwrites existing files**

Rule: `CORRECTNESS-001`

Every export for a slot writes to the same home-directory path with default overwrite behavior. Exporting another certificate for that slot, or encountering an unrelated file with that name, destroys the existing contents without confirmation.

**Suggested resolution:** Use a user-selected destination or a collision-resistant filename, and refuse or confirm before overwriting an existing file.

clankus-aurelius (approved):
Three actionable issues remain: cancellation can leave hidden smart-card work running, failed clipboard-blob deletion is reported as successful, and certificate export silently overwrites existing files.

---

clankus-aurelius:
@aurelleb automated review passed for `f887287`; this extension is ready for your review.

LLawli:
All three fixed in `bc70db4`.

**`src/lib/touch.ts` — the cancellation race.** Correct, and I missed it when I added
`abort()` last round: the abort was reachable only through `onCancel`, which is assigned
after `conn.connect()` resolves, so a cancel landing in that window had nothing to call
and the touch CALCULATE went out regardless. Cancellation now sets its own flag, checked
before the transaction starts, and `waitForCard()` got the same treatment. Verified
against the key: an immediate cancel rejects at 0ms and the card answers a follow-up
request 57ms later, instead of staying locked for the timeout.

**`src/lib/clipboard.ts` — a surviving blob is not a purge.** Also correct. The selection
row is only the index; the code itself is the file under `clipboard-data`. A failed
`rmSync` now makes the function return `failed`, which the HUD added in an earlier round
already surfaces.

**`src/security-keys.tsx` — no more silent overwrite.** The write uses `wx` and, when the
name is taken, falls back to one carrying the certificate's fingerprint; re-exporting the
same certificate lands on the same file and reports it was already exported. I kept the
home directory rather than adding a destination picker, since `@vicinae/api` exposes no
save dialog that I could find — happy to change the approach if there is one I missed.
Verified: exporting a second, different certificate for the same slot leaves the first
file byte-for-byte intact.

One caveat on the evidence, so it is not overstated: the clipboard fix is the one change
here I could not exercise end to end, since forcing a blob-removal failure means
interfering with a live Vicinae clipboard database. It is a one-line propagation of an
already-logged failure, reviewed by reading.

extensions/yubikey/src/security-keys.tsx
```diff
@@ -0,0 +1,438 @@
...
+  });
+
+  // The PIN lives only here, in the command's memory. Never in Cache, LocalStorage or argv.
+  const [fidoPin, setFidoPin] = useState<string | null>(null);
+  const [fidoCreds, setFidoCreds] = useState<FidoCred[] | null>(null);
+
+  const load = useCallback(async () => {
+    // The two halves are independent: PIV rides on CCID (pcscd), FIDO2 on its own HID
+    // interface. A machine with no pcscd still manages passkeys, so neither failure is
+    // allowed to take the other section down with it.
```
clankus-aurelius, line 53: **🟠 Warning — Reload retains credentials and PIN from the previous key**

Rule: `CORRECTNESS-001`

load() refreshes device information without clearing fidoCreds or fidoPin. After swapping the sole connected YubiKey and reloading, the screen still presents the previous key's passkeys and deletion attempts use the old PIN and credential ID against the new key.

**Suggested resolution:** Clear the cached passkey list and in-memory PIN whenever device state is reloaded.

```suggestion
const load = useCallback(async () => {
    setFidoPin(null);
    setFidoCreds(null);
```

extensions/yubikey/src/lib/piv.ts
```diff
@@ -0,0 +1,260 @@
...
+  const inner = parseTlvs(outer.value);
+  const certTlv = inner.find((x) => x.tag === TAG.CERTIFICATE);
+  const infoTlv = inner.find((x) => x.tag === TAG.CERT_INFO);
+  if (!certTlv) return null;
+
+  let der = certTlv.value;
+  const certInfo = infoTlv && infoTlv.value.length > 0 ? infoTlv.value[0] : 0;
+  if (certInfo === 1) der = gunzipSync(der); // compressed certificate
+
+  const x = new X509Certificate(der);
```
clankus-aurelius, line 183: **🟠 Warning — Certificate listing misreads the compression flag**

Rule: `CORRECTNESS-001`

The documented PIV contract defines compression as certInfo & 0x01, but this equality check only decompresses when the entire byte is exactly 1. Certificates with compression plus another flag fail parsing.

**Suggested resolution:** Test the compression bit rather than the complete metadata byte.

```suggestion
if ((certInfo & 0x01) !== 0) der = gunzipSync(der); // compressed certificate
```

extensions/yubikey/src/lib/piv.ts
```diff
@@ -0,0 +1,260 @@
...
+  const outer = parseTlvs(data).find((x) => x.tag === TAG.OBJ_DATA);
+  const inner = outer ? parseTlvs(outer.value) : [];
+  const certTlv = inner.find((x) => x.tag === TAG.CERTIFICATE);
+  const infoTlv = inner.find((x) => x.tag === TAG.CERT_INFO);
+  if (!certTlv) throw new PivError("no_cert", "There is no certificate in that slot");
+
+  let der = certTlv.value;
+  if (infoTlv && infoTlv.value.length > 0 && infoTlv.value[0] === 1) der = gunzipSync(der);
+
+  return new X509Certificate(der).toString(); // PEM
```
clankus-aurelius, line 209: **🟠 Warning — Certificate export misreads the compression flag**

Rule: `CORRECTNESS-001`

The export path repeats the exact-value comparison, so a certificate whose compression bit is combined with another certInfo flag cannot be exported.

**Suggested resolution:** Test bit 0 of certInfo before decompressing.

```suggestion
if (infoTlv && infoTlv.value.length > 0 && (infoTlv.value[0] & 0x01) !== 0) der = gunzipSync(der);
```

extensions/yubikey/src/lib/i18n.ts
```diff
@@ -0,0 +1,204 @@
...
+  "err.touch_timeout": "The YubiKey was not touched in time",
+  "err.no_applet": "This key does not expose the application. CCID may be disabled on it.",
+  "err.pin_invalid": "Wrong PIV PIN. {retries} attempts left.",
+  "err.pin_blocked": "PIV PIN blocked. The PUK is required to unblock it.",
+  "err.no_cert": "There is no certificate in that slot",
+  "err.no_device": "No YubiKey FIDO interface found",
+  "err.no_access": "No permission to access the YubiKey over HID. Install the FIDO udev rules (`libfido2` on Fedora/Arch, `libu2f-udev` on Debian/Ubuntu).",
+  "err.ctap.49": "Wrong PIN",
+  "err.ctap.52": "Too many attempts in a row. Remove the YubiKey and reinsert it.",
+  "err.ctap.53": "PIN blocked. Remove the YubiKey and reinsert it.",
```
clankus-aurelius, line 201: **🟠 Warning — PIN_NOT_SET is reported as a blocked PIN**

Rule: `CORRECTNESS-001`

CTAP status 0x35 is PIN_NOT_SET according to the supplied protocol reference, but decimal status 53 tells users the PIN is blocked and recommends reinserting the key, which cannot resolve the failure.

**Suggested resolution:** Explain that no FIDO2 PIN is configured and direct the user to configure one.

```suggestion
"err.ctap.53": "No FIDO2 PIN is set on this YubiKey.",
```

extensions/yubikey/src/lib/pcsc.ts
```diff
@@ -0,0 +1,562 @@
...
+    if (existsSync(path)) return path;
+  }
+
+  throw new PcscError(
+    "no_daemon",
+    "pcscd is not running. Install `pcsc-lite` (or `pcscd` on Debian/Ubuntu) and enable it with: " +
+      "sudo systemctl enable --now pcscd.socket",
+  );
+}
+
```
clankus-aurelius, line 151: **🟠 Warning — Invalid socket preference silently falls back to another socket**

Rule: `CORRECTNESS-001`

The configured override is merely the first existing candidate. If the user-specified path is missing, socketPath silently selects a system default, so operations can run against a daemon other than the one explicitly configured.

**Suggested resolution:** When a non-empty override is configured, use only that path and surface a clear connection or missing-path error; consult defaults only when the preference is empty.

clankus-aurelius (approved):
Found five actionable correctness issues affecting FIDO2 state, PIV certificate handling, error guidance, and custom pcscd configuration.

---

clankus-aurelius:
@aurelleb automated review passed for `bc70db4`; this extension is ready for your review.

PR #332, opened 2026-07-17
Add Snapper Snapshots extension

samvoelzke:
Adds **Snapper Snapshots** — browse, create and inspect Btrfs snapshots managed by Snapper (the engine behind Btrfs Assistant) from Vicinae.

## Commands
- **Browse Snapshots** — list Snapper configs and snapshots, with a decoded, readable changed-files view and on-demand per-file diffs.
- **Search Snapshot Files** — instant local file search inside any snapshot's read-only mount.
- **Create Snapshot** — take a new snapshot with a description.

## First run
Snapper only allows root to read snapshots by default, so the first open shows a one-tap **Set Up Access** screen that adds your user to `ALLOW_USERS` + `SYNC_ACL` via a single `pkexec` prompt. After that everything works without further prompts.

Everything runs locally — no network. Full subvolume rollbacks are deliberately left to Btrfs Assistant.


aurelleb:
@clankus-aurelius review

clankus-aurelius:

🔴 **Contributor changes requested.** Address the blocking inline findings and push a new commit; the bot will review it automatically.

4 blocking findings must be addressed.

extensions/snapper/src/browse-snapshots.tsx
```diff
@@ -0,0 +1,396 @@
...
+  const { data, isFirstLoad, error } = useCached<string>(
+    `diff:${config.config}:${snapshot.number}:${path}`,
+    () => getDiff(config.config, snapshot.number, 0, path),
+  );
+  const name = path.split("/").pop() || path;
+
+  let md: string;
+  if (error && !data) {
+    md = `# ${name}\n\n${error}`;
+  } else if (isFirstLoad && !data) {
```
clankus-aurelius, line 312: **🔴 Blocking — File diffs are persisted in Cache**

Rule: `SECURITY-003`

`useCached` writes every result to `Cache`, so arbitrary snapshot file contents returned by `getDiff` are persisted outside the view lifecycle and may include secrets.

**Suggested resolution:** Keep diff contents in component memory without persistent caching. If persistence is essential, use `LocalStorage` from `@vicinae/api` and make the storage behavior explicit.

extensions/snapper/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "license": "MIT",
+  "platforms": ["Linux"],
+  "author": "samvoelzke",
+  "contributors": [],
+  "icon": "extension_icon.png",
+  "commands": [
+    {
+      "name": "browse-snapshots",
+      "title": "Browse Snapshots",
+      "subtitle": "Snapper",
```
clankus-aurelius, line 16: **🔴 Blocking — Manifest claims unsupported restore functionality**

Rule: `MANIFEST-001`

The extension and Browse command descriptions promise snapshot/file restoration, but the implementation only opens snapshot files and explicitly leaves rollback to Btrfs Assistant.

**Suggested resolution:** Describe the implemented browse, inspect, create, search, and open functionality without claiming restoration, or implement the advertised restore actions.

extensions/snapper/src/browse-snapshots.tsx
```diff
@@ -0,0 +1,396 @@
...
+                <Action
+                  title="Reveal in File Manager"
+                  icon={Icon.Folder}
+                  onAction={() =>
+                    open(
+                      (snapshotMountPath(config, snapshot.number) + c.path).replace(/\/[^/]*$/, "") ||
+                        "/",
+                    )
+                  }
+                />
```
clankus-aurelius, line 279: **🔴 Blocking — File reveal reimplements a portable Vicinae API**

Rule: `API-001`

The action manually derives a parent directory and calls `open`, while current `@vicinae/api` provides `showInFileBrowser` for this complete use case.

**Suggested resolution:** Import and call `showInFileBrowser` with the snapshot file path. Upgrade `@vicinae/api` to 0.24.0 if needed, run `npm install`, and commit the updated lockfile.

extensions/snapper/src/lib/snapper.ts
```diff
@@ -0,0 +1,258 @@
...
+  return data.configs ?? [];
+}
+
+/** True if the current user can read snapshots (ALLOW_USERS granted) for at least one config. */
+export async function hasAccess(): Promise<boolean> {
+  let configs: SnapperConfig[];
+  try {
+    configs = await listConfigs();
+  } catch {
+    return false;
```
clankus-aurelius, line 79: **🟠 Warning — Missing Snapper is presented as an access problem**

Rule: `UX-001`

Every `listConfigs` failure, including a missing `snapper` executable, becomes `false`, sending users to a privileged access setup that cannot solve the failure.

**Suggested resolution:** Return `false` only for confirmed permission errors; propagate missing-command and other operational errors so the UI can explain how to install or repair Snapper.

extensions/snapper/src/lib/snapper.ts
```diff
@@ -0,0 +1,258 @@
...
+  if (configs.length === 0) return true; // nothing to gate on
+  try {
+    await listSnapshots(configs[0].config);
+    return true;
+  } catch (e) {
+    if (e instanceof NoAccessError) return false;
+    // A different error (e.g. transient) shouldn't force the setup screen.
+    return true;
+  }
+}
```
clankus-aurelius, line 90: **🟠 Warning — Access is checked for only one config**

Rule: `CORRECTNESS-001`

Access to the first config returns `true` even when later configs remain inaccessible, so the advertised setup gate is skipped and those configs fail when opened.

**Suggested resolution:** Check every discovered config and return `false` if any snapshot listing raises `NoAccessError`; propagate unrelated failures.

extensions/snapper/src/search-files.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+    const id = ++runId.current;
+    setLoading(true);
+    const handle = setTimeout(async () => {
+      try {
+        const res = await searchInSnapshot(target.mountPath, query.trim());
+        if (id === runId.current) setHits(res);
+      } catch {
+        if (id === runId.current) setHits([]);
+      } finally {
```
clankus-aurelius, line 83: **🟠 Warning — Snapshot search failures are shown as no matches**

Rule: `UX-001`

All `find` failures are silently converted to an empty result, making missing tools, permission failures, and I/O errors indistinguishable from a successful search with no matches.

**Suggested resolution:** Store and display the caught error, while reserving the empty state for successful searches with zero results.

extensions/snapper/src/browse-snapshots.tsx
```diff
@@ -0,0 +1,396 @@
...
+                    icon={Icon.BulletPoints}
+                    target={<DiffView config={config} snapshot={snapshot} path={c.path} />}
+                  />
+                )}
+                <Action
+                  title="Open Snapshot Version"
+                  icon={Icon.Eye}
+                  onAction={() => open(snapshotMountPath(config, snapshot.number) + c.path)}
+                />
+                <Action
```
clankus-aurelius, line 270: **🟠 Warning — Created files offer a nonexistent snapshot version**

Rule: `CORRECTNESS-001`

For a snapshot-to-current comparison, `+` means the file exists only in the current system, but the unconditional action tries to open it inside the older snapshot.

**Suggested resolution:** Show “Open Snapshot Version” only when `c.status[0] !== "+"`, mirroring the existing current-version guard for deleted files.

extensions/snapper/src/lib/snapper.ts
```diff
@@ -0,0 +1,258 @@
...
+    });
+    child.on("error", reject);
+  });
+}
+
+export function launchBtrfsAssistant(): void {
+  execFile("btrfs-assistant-launcher", (err) => {
+    if (err) execFile("btrfs-assistant", () => undefined);
+  });
+}
```
clankus-aurelius, line 258: **🟠 Warning — Btrfs Assistant launch failures are silent**

Rule: `UX-001`

If neither launcher executable exists or either launch fails, the action produces no user-visible feedback.

**Suggested resolution:** Report the final launch error with a failure toast or HUD and explain that Btrfs Assistant is optional but must be installed for this action.

extensions/snapper/src/create-snapshot.tsx
```diff
@@ -0,0 +1,77 @@
...
+    >
+      <Form.Dropdown id="config" title="Config">
+        {configs.map((c) => (
+          <Form.Dropdown.Item key={c.config} value={c.config} title={`${c.config} (${c.subvolume})`} />
+        ))}
+      </Form.Dropdown>
+      <Form.TextField id="description" title="Description" placeholder="e.g. before system update" />
+    </Form>
+  );
+}
```
clankus-aurelius, line 63: **🟠 Warning — Create command has no empty-config state**

Rule: `UX-002`

When Snapper has no configs, the form shows an empty dropdown but still exposes submission, which passes an undefined config and fails instead of explaining that setup is required.

**Suggested resolution:** Render an actionable empty state when `configs` is empty and prevent submission until a valid config is selected.

clankus-aurelius (changes requested):
The extension has publication blockers around privileged shell construction, sensitive diff caching, misleading restore claims, and bypassing the portable file-reveal API, plus several workflow defects.

---
_Automated review found 4 publication-blocking issues._

PR #333, opened 2026-07-18
feat: seerr extension

vMohammad24:
# Seerr extension

An extension to discover and request movies or TV shows from A [Seerr Instance](https://github.com/seerr-team/seerr)


aurelleb:
looks like you need to regenerate the lock file

PR #334, opened 2026-07-18
refactor(seerr): change author username

vMohammad24:


aurelleb:
this will remain broken until I'm brave enough to fix it internally lol

PR #338, opened 2026-07-21
add rbw extension

bl4zee1g:
this pr adds an extension that lets you browse your bitwarden vault through [rbw](https://github.com/doy/rbw) also i added a kaomoji to my other extension, hope that's aighty 𑣲⋆


extensions/bitwarden/src/rbw.ts
```diff
@@ -0,0 +1,99 @@
...
+    const err = error as NodeJS.ErrnoException & { stderr?: string };
+    throw new RbwError(
+      `rbw command failed: ${err.message}`,
+      typeof err.stderr === "string" ? err.stderr : undefined,
+    );
+  }
+}
+
+/** Check if the vault is currently unlocked. */
+export async function isUnlocked(): Promise<boolean> {
```
aurelleb, line 58: I think we want a proper error message for when the `rbw` binary is not in PATH, so that it's clear to the user they don't have the required CLI for it to work. Currently it is getting conflated with a locking problem.

extensions/rbw/package.json
```diff
@@ -0,0 +1,41 @@
...
+	"$schema": "https://raw.githubusercontent.com/vicinaehq/vicinae/refs/heads/main/extra/schemas/extension.json",
+	"name": "rbw",
+	"title": "rbw",
+	"description": "Search, copy and paste credentials from your Bitwarden vault using the rbw CLI",
+	"categories": [
+		"Security"
+	],
+	"license": "MIT",
+	"platforms": [
+		"Linux"
```
aurelleb: We can probably add macOS to the list, I don't think this does anything Linux specific?

extensions/rbw/package.json
```diff
@@ -0,0 +1,41 @@
+{
+	"$schema": "https://raw.githubusercontent.com/vicinaehq/vicinae/refs/heads/main/extra/schemas/extension.json",
+	"name": "rbw",
```
aurelleb: worth renaming `bitwarden`: we won't have multiple bitwarden extensions in the store anyways.

extensions/rbw/package.json
```diff
@@ -0,0 +1,41 @@
...
+	"title": "rbw",
```
aurelleb: `Bitwarden (rbw)` would make for a better title

extensions/kaomojis/src/kaomojis.ts
```diff
@@ -1,4 +1,5 @@
 export const kaomojis: string[] = [
+	"₍^˶ ╸𖥦 ╸˵^₎⟆",
```
aurelleb, line 2: although that's a very tiny change, next time make a separate PR for it, we try to keep it so that one PR only affects one extension. That's fine for this time, no need to do it

extensions/rbw/src/browse-vault.tsx
```diff
@@ -0,0 +1,504 @@
...
+  return { name, user: user ?? null, password, fields, notes: notes ?? null };
+}
+
+async function performAction(
+  title: string,
+  value: string,
+  actionType: "copy" | "paste" = "copy",
+): Promise<void> {
+  if (actionType === "copy") {
+    await Clipboard.copy(value);
```
aurelleb: We probably want to pass `{ concealed: true }` in order to exclude this from the clipboard history, no?
Maybe you can make it a preference in the future.

aurelleb (changes requested):
I really like the simple approach you took here, and I'm not willing to review the two other bitwarden extensions that are full of slop, so I think we will go with this one as a base.

I guess in the future we could try to make it multi backend and all that, but for now it does its main job correctly.

Just a few things to change before we can merge.

aurelleb:
Also, it looks like the vault unlock is failing: it shows a toast and closes the window immediately (difficult to see). I can unlock the vault normally using `rbw unlock`. It would be great to use a [Form view](https://docs.vicinae.com/extensions/api/form) to handle this.
This needs to be addressed before merge

bl4zee1g:
All feedback except for your last point about the unlock failing I addressed in the latest push. I have tried using the Form view to handle unlocking but couldn't get it to work nicely since rbw doesn't support something like rbw unlock [vault password] for security reasons, it enforces the use of pinentry as far as I know. could you check what that toast message you get when clicking unlock in the extension is about? because for me it just spawns a pinentry window same exact way simply running rbw unlock would. If it's really needed I'll spend some more time reading through pinentry docs I just thought since this only supports rbw for now it's enough to have it be like this. Also just a heads up I won't be able to work on this or respond to feedback for the next week since I'll be away.

aurelleb:
My bad I had my pinentry misconfigured, it works great

PR #343, opened 2026-07-28
feat: add prompt-deck extension

tadassuksteris:
## Summary

PromptDeck lets you define reusable AI prompts and run them against your selected or
clipboard text. Each prompt can be registered as a Quicklink, so it becomes its own
entry in root search.

The Manage Prompts list, showing each prompt's run mode and context sources

A PromptDeck prompt found in Vicinae root search

## Commands

- **Manage Prompts** — create, edit and run prompts
- **Run Prompt** — runs a saved prompt by id; used by the Quicklinks it creates
- **Prompt History** — browse, re-run and copy past runs; can be disabled entirely

## Providers and network access

The only remote hosts contacted are the LLM endpoints the user configures:
`api.openai.com`, `api.anthropic.com`, `generativelanguage.googleapis.com`, or a
custom OpenAI-compatible base URL such as Ollama or OpenRouter. No other network
calls are made, and nothing is contacted until the user runs a prompt.

What gets sent is the prompt itself, plus selected and clipboard text when a prompt
opts into them. This is stated in the manifest description and the README's Privacy
section. API keys are stored as Vicinae password preferences, one per provider;
local endpoints that need no auth work with the key left blank. Run history is
stored locally through LocalStorage and can be turned off.

Tested against OpenAI, Google, and an OpenAI-compatible endpoint (OpenRouter).

## Notes

`media/` holds README images only. `vici build` bundles `assets/`, so they are not
shipped to users.


aurelleb:
Pretty cool, I tested with with haiku 4.5 and it works well.
Congratulations for your first extension!

PR #359, opened 2026-08-19
feat: add buku extension

msclp:
Adds an integration for [buku](https://github.com/jarun/buku), the command line
bookmark manager. Bookmarks are read from and written to your existing buku database -
the extension keeps no copy of its own.

## Commands

- **Search Bookmarks** - fuzzy search across titles, URLs, and tags, with tags shown as
  row accessories. Open in browser, copy the URL, edit, delete, or add a new bookmark.
  A toggleable detail pane (`Ctrl+Shift+D`) surfaces the description, which has nowhere
  to go on a single row, along with the URL, tags, and buku ID. Refresh (`Ctrl+R`)
  re-reads the database after changes made with the `buku` CLI itself.
- **Add Bookmark** - prefills the URL from the clipboard when it holds one. Leaving the
  title empty is buku's cue to fetch it from the page; typing one skips the fetch. The
  form stays open after saving so several bookmarks can be added in a row.

Both forms take a comma separated tag list, and `Ctrl+T` opens a picker over the tags
already in the database, with toggling and the option to create a tag from the search
text. Tags are lowercased and de-duplicated before they reach buku, matching how buku
stores them.

## Requirements

- The `buku` command line tool on `PATH` (`paru -S buku`, `pipx install buku`, …)

Vicinae runs extensions with the desktop session's `PATH`, which frequently does not
include `~/.local/bin`, so a `pipx`/`pip --user` install of buku is invisible to it. The
extension therefore also looks in `~/.local/bin`, `/usr/local/bin`, and
`/opt/homebrew/bin`, and offers an **Additional PATH Entries** preference for anything
else. When buku still cannot be found, the empty view says so and links to the
preferences rather than surfacing `spawn buku ENOENT`.

## Notes

- **`Form.TagPicker` is not used**, because it renders nothing: the server parses
  `tag-picker-field` and then discards it (`FormTagPickerFieldWire` maps to an empty
  handler in `model-deser.cpp`, and `FormModel::Field` has no tag picker variant - same
  on `main`). Tags are picked in a pushed `List` instead. The selection lives inside that
  pushed view rather than in the form, since `push(target)` snapshots the element and it
  never sees updated props.
- **Clearing every tag needs an explicit removal.** buku ignores `--tag ""`
  (`update_rec` turns an empty string into `None`), so wiping the last tag off a bookmark
  is issued as `--tag - "old,tags"`, with the `-` as its own argument and the tags as a
  single comma separated one - otherwise buku stores `-`, `old` and `tags` as literal tag
  names. `--tag` is also kept last in the argument list because it takes a variable number
  of values and only stops at the next flag.
- **buku's title fetch has no timeout of its own**, so an unreachable host would hang the
  add until the extension cuts it off. That case is reported as a timeout with a hint to
  fill in the title, rather than as a generic failure.
- `buku -p -j` output is scanned for the start of the JSON array, because buku prefixes it
  with notices such as the one printed the first time it creates the database.
- Only one dependency: `@vicinae/api`.

## Icon

The 🔖 glyph from [Noto Emoji](https://github.com/googlefonts/noto-emoji), Copyright 2013
Google LLC, Apache-2.0. Attributed in the extension's README.


clankus-aurelius:

extensions/buku/src/components/bookmark-empty-state.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <List.EmptyView
+        icon={Icon.XMarkCircle}
+        title={missing ? "buku not found" : "Could not read bookmarks"}
+        description={missing ? BUKU_MISSING_HINT : describeError(error)}
+        actions={
+          <ActionPanel>
+            {reload}
+            <Action title="Open Extension Preferences" icon={Icon.Cog} onAction={() => openCommandPreferences()} />
+          </ActionPanel>
+        }
```
clankus-aurelius, line 43: **🟠 Warning — Missing-buku recovery opens the wrong preferences**

Rule: `CORRECTNESS-001`

“Additional PATH Entries” is an extension-level preference, but this action calls `openCommandPreferences()`. The API declarations distinguish this from `openExtensionPreferences()`, so users are sent to settings that do not contain the advertised recovery option.

**Suggested resolution:** Import `openExtensionPreferences` instead of `openCommandPreferences` and invoke it from this action.

clankus-aurelius (approved):
One actionable recovery-flow issue was found.

---

clankus-aurelius:
@aurelleb automated review passed for `c8ca23b`; this extension is ready for your review.

clankus-aurelius (approved):
The previous recovery-flow issue is resolved, and the incremental change introduces no actionable problems.

---

clankus-aurelius (approved):
The incremental preference-opening change introduces no actionable issues.

---

aurelleb:
thanks!

PR #366, opened 2026-08-26
fix(vscode-recents): make it work on macOS

### LINKED ISSUE #365: [vscode-recents] Broken on macOS: wrong shared storage path (empty list) and unreachable executable
### Extension name

vscode-recents

### Description

On macOS the extension opens and shows **"No Recent Projects Found"**, with no error, while VS Code itself still lists its recent projects. If anything does show up, pressing Enter produces a **"Command not found"** toast and nothing opens.

This is the same symptom as #264, but a different cause: the fix in #272 only produced the correct path on Linux. I dug into it on macOS and found four distinct bugs. I've opened a PR for three of them plus a cosmetic one; the fourth (WAL) is described below but left out, since it overlaps with #202.

---

#### 1. Shared storage path is wrong on macOS and Windows — this is what makes the list empty

`VSCODE_SHARED_STATE_PATHS` derives the path per platform:

```ts
darwin: (home, flavour) => `${home}/Library/Application Support/${sharedDir}/sharedStorage/state.vscdb`
win32:  (home, flavour) => `${home}/AppData/Roaming/${sharedDir}/sharedStorage/state.vscdb`
```

There is no such per-platform variation. VS Code resolves it as:

```js
get appSharedDataHome() {
    const dir = this.args["shared-data-dir"];
    if (dir) return URI.file(resolve(dir));
    const portable = process.env.VSCODE_PORTABLE;
    return portable ? URI.file(join(portable, "shared-data"))
                    : joinPath(this.userHome, this.productService.sharedDataFolderName);
}
```

`userHome` is plain `os.homedir()`, so the directory is `~/<sharedDataFolderName>` on Linux, macOS **and** Windows. Only the Linux branch was ever right.

On macOS:

```console
$ ls ~/.vscode-shared/sharedStorage/
state.vscdb  state.vscdb-shm  state.vscdb-wal  state.vscdb.backup

$ ls ~/Library/Application\ Support/.vscode-shared
ls: no such file or directory
```

**Why it fails silently:** since the shared path doesn't exist, the lookup falls back to `~/Library/Application Support/Code/User/globalStorage/state.vscdb`. That file *does* still exist (322 keys here), so the "database not found" error never fires — but it no longer holds `history.recentlyOpenedPathsList`. Empty list, no diagnostic.

> Sidenote, not in the PR: reading `sharedDataFolderName` straight out of the app bundle's `product.json` instead of hand-maintaining `VSCODE_SHARED_STORAGE_DIRS` would have avoided the VSCodium mismatch that needed #304, and would keep working for flavours nobody has mapped yet.

#### 2. `command -v` can never succeed on macOS — this is the "Command not found"

The Vicinae server is started by LaunchServices, not from a login shell:

```console
$ ps eww -p $(pgrep -f Vicinae | head -1) | tr ' ' '\n' | grep ^PATH=
PATH=/usr/bin:/bin:/usr/sbin:/sbin
```

The extension runtime inherits that. The VS Code CLI is not on it:

```console
$ ls -l /usr/local/bin/code
/usr/local/bin/code -> /Applications/Visual Studio Code.app/Contents/Resources/app/bin/code

$ env -i PATH=/usr/bin:/bin:/usr/sbin:/sbin sh -c 'command -v code'; echo $?
1
```

So `isExecutableAvailable()` returns `false` every time. Homebrew casks (`/opt/homebrew/bin`) are out of reach too, and since `exec()` runs a non-interactive `/bin/sh`, having `code` in the user's `.zshrc` doesn't help either.

#### 3. `--folder-uri` is handed a path, not a URI

`decodeFileUri()` strips `file://` and percent-decodes; the result goes straight into `--folder-uri`, which parses its argument as a URI. Simple paths survive, but `#` and `?` truncate:

```console
$ code --new-window --folder-uri "/tmp/vsr proyecto#uno"
# opens an empty window titled "vsr proyecto", not "vsr proyecto#uno"
```

#### 4. The WAL is never read — stale list, and unsafe writes *(not in the PR)*

VS Code keeps the DB in WAL mode; recent writes live in `state.vscdb-wal` until SQLite checkpoints. `sql.js` gets a flat buffer of `state.vscdb` alone and cannot see them. Measured here with VS Code running:

| read | entries | first project |
| --- | --- | --- |
| `state.vscdb` only (what sql.js does) | 580 | a stale remote one |
| with the WAL applied (reality) | 585 | the one opened moments earlier |

Worse, `removeRecentProject()` does `db.export()` of that stale snapshot and writes it back **over** `state.vscdb` while VS Code holds the file open, discarding every pending WAL frame.

Worth knowing: `node:sqlite` is available **unflagged** in the Vicinae extension runtime (Node v22.23.1). It reads through the WAL and writes under proper SQLite locking, and would drop the `sql.js` dependency and its `.wasm` asset entirely. I left this out of the PR because #202 is already reworking the database layer — happy to send it separately if that's useful.

#### 5. Cosmetic: duplicated workspace suffix

`getProjectLabel()` appends `.code-workspace` without checking whether the basename already ends in it, producing `MyProject.code-workspace.code-workspace`.

### Steps to reproduce

1. macOS with VS Code 1.118 or newer (tested on 1.134.0), opened at least once so recents exist.
2. Set the "VSCode Flavour" preference to *Visual Studio Code*.
3. Run **Open Recents** → "No Recent Projects Found".
4. Confirm the data is actually there: `sqlite3 ~/.vscode-shared/sharedStorage/state.vscdb "select value from ItemTable where key = 'history.recentlyOpenedPathsList';" | jq '.entries | length'`
5. For bug 2, temporarily hardcode the correct path, then press Enter on any entry → "Command not found".

### Vicinae version

0.27.0

### Additional context

macOS 15 (Darwin 25.5.0) · VS Code 1.134.0 · extension at 9fca13cd3839c5046d7099d4a2263d8dc8263b74

After fixing 1–3 locally the command lists 293 valid projects (including SSH remotes) and opens them correctly under the server's restricted PATH.

Related: #264 (same symptom, Linux), #272 (fixed Linux only), #304 (VSCodium mapping), #202 (open, database layer).

tonicastillo:
Fixes #365.

On macOS the extension shows an empty list, and anything that does show up fails to open with "Command not found". Three separate macOS bugs, plus one cosmetic fix. Full evidence is in the issue; the short version:

### 1. Shared storage path (`src/constants.ts`, `src/helpers.ts`)

The shared storage directory hangs directly off the home directory on **every** platform, not under the per-platform application data directory — VS Code resolves it as `joinPath(userHome, product.sharedDataFolderName)`, and `userHome` is plain `os.homedir()`. The per-platform table added in #272 only produced the right path on Linux.

On macOS the lookup then fell through to the legacy per-flavour DB, which still exists but no longer holds `history.recentlyOpenedPathsList` — so the "database not found" error never fired and the list just came up empty with no diagnostic. The error message now names both paths that were tried, so the next person to hit this gets something to work with.

`VSCODE_SHARED_STATE_PATHS` is replaced by a single `getSharedStateDBPath()`.

### 2. Executable lookup (`src/util/vscode.ts`)

The Vicinae server is launched by LaunchServices, not from a login shell, so extensions inherit `PATH=/usr/bin:/bin:/usr/sbin:/sbin`. The `code` shim lives in `/usr/local/bin` and Homebrew casks use `/opt/homebrew/bin`, so `command -v code` failed every single time and nothing could ever be opened. `exec()` also runs a non-interactive `/bin/sh`, so the user's shell profile doesn't help.

On darwin the search path is now widened, and if the CLI still isn't found we fall back to the one inside the application bundle (`/Applications/<App>.app/Contents/Resources/app/bin/<cli>`), which is there even when the user never ran *Shell Command: Install 'code' command in PATH*. The widened path is passed to the final `exec` too. Non-darwin behaviour is unchanged.

### 3. `--folder-uri` was handed a path, not a URI (`src/util/vscode.ts`)

`decodeFileUri()` strips `file://` and percent-decodes, and the result went straight into `--folder-uri`, which parses its argument as a URI. Plain paths survive, but `#` and `?` truncate — `/tmp/vsr proyecto#uno` silently opened an empty `vsr proyecto` window. Local paths are re-encoded into a proper `file://` URI; anything that already carries a scheme (`vscode-remote://…`) passes through untouched.

### 4. Cosmetic (`src/helpers.ts`)

`getProjectLabel()` appended `.code-workspace` without checking whether the basename already ended in it, giving `MyProject.code-workspace.code-workspace`.

### Deliberately not included

The WAL bug (item 4 in the issue): `sql.js` reads a flat buffer of `state.vscdb` and never sees `state.vscdb-wal`, so the list is stale, and `removeRecentProject()` writes that stale snapshot back over the live DB. It's a real bug, but #202 is already reworking the database layer and I'd rather not collide with it. Happy to send it separately — `node:sqlite` is available unflagged in the extension runtime (Node v22.23.1), reads through the WAL, and would let `sql.js` and its `.wasm` asset go.

### Testing

macOS 15 (Darwin 25.5.0), Vicinae v0.27.0, VS Code 1.134.0.

- Lists 293 valid projects, SSH remotes included; before the fix, zero.
- Opens correctly under the server's exact restricted PATH (`env -i PATH=/usr/bin:/bin:/usr/sbin:/sbin`), verified against both a `code` on PATH and the bundled CLI.
- `/tmp/vsr proyecto#uno` now opens that folder, with its contents, instead of a truncated empty one.
- `npx vici build` clean; `biome format` reports no changes; `package-lock.json` untouched.

I don't have Linux or Windows to hand, so those paths deserve a second pair of eyes — though bug 1 should be a no-op on Linux (same string as before) and bugs 2 and 4 are guarded by `process.platform !== "darwin"`.

One style note: a few existing functions became arrow functions, which widens the diff more than the change warrants. Say the word and I'll put the declarations back.

### DIFF
--- a/extensions/vscode-recents/package.json
+++ b/extensions/vscode-recents/package.json
@@ -5,7 +5,7 @@
     "description": "Open VSCode and compatible editors recent projects",
     "license": "MIT",
     "platforms": ["Linux", "macOS"],
-    "contributors": [],
+    "contributors": ["tonicastillo"],
     "pastContributors": [],
     "author": "ShyAssassin",
     "icon": "extension_icon.png",
--- a/extensions/vscode-recents/src/constants.ts
+++ b/extensions/vscode-recents/src/constants.ts
@@ -28,23 +28,42 @@ const VSCODE_SHARED_STORAGE_DIRS: Record<VSCodeFlavour, string> = {
     [VSCodeFlavour.CodeInsiders]: ".vscode-insiders-shared",
 };
 
-export const VSCODE_SHARED_STATE_PATHS = {
-    linux: (home: string, flavour: VSCodeFlavour) => {
-        const sharedDir = VSCODE_SHARED_STORAGE_DIRS[flavour];
-        return `${home}/${sharedDir}/sharedStorage/state.vscdb`;
-    },
-    win32: (home: string, flavour: VSCodeFlavour) => {
-        const sharedDir = VSCODE_SHARED_STORAGE_DIRS[flavour];
-        return `${home}/AppData/Roaming/${sharedDir}/sharedStorage/state.vscdb`;
-    },
-    darwin: (home: string, flavour: VSCodeFlavour) => {
-        const sharedDir = VSCODE_SHARED_STORAGE_DIRS[flavour];
-        return `${home}/Library/Application Support/${sharedDir}/sharedStorage/state.vscdb`;
-    },
-} as const;
+// FIX(macOS/Windows): the shared storage directory hangs directly off the user's
+// home directory on EVERY platform, not under the per-platform application data
+// directory. VS Code resolves it as:
+//     get appSharedDataHome() { ... return joinPath(this.userHome, this.productService.sharedDataFolderName) }
+// (vs/platform/environment/node/environmentService.ts), and `userHome` is plain
+// `os.homedir()`. The previous per-platform table only produced the right path on
+// Linux; on macOS it pointed at "~/Library/Application Support/.vscode-shared",
+// which never exists, so the lookup silently fell back to the legacy DB - a file
+// that still exists but no longer holds `history.recentlyOpenedPathsList`,
+// yielding an empty list with no error.
+export const getSharedStateDBPath = (home: string, flavour: VSCodeFlavour): string => {
+    const sharedDir = VSCODE_SHARED_STORAGE_DIRS[flavour];
+    return `${home}/${sharedDir}/sharedStorage/state.vscdb`;
+};
 
 export const VSCODE_STATE_PATHS = {
     linux: (home: string, flavour: VSCodeFlavour) => `${home}/.config/${flavour}/User/globalStorage/state.vscdb`,
     win32: (home: string, flavour: VSCodeFlavour) => `${home}/AppData/Roaming/${flavour}/User/globalStorage/state.vscdb`,
     darwin: (home: string, flavour: VSCodeFlavour) => `${home}/Library/Application Support/${flavour}/User/globalStorage/state.vscdb`,
 } as const;
+
+// FIX(macOS): the Vicinae server is started by LaunchServices, so extensions inherit
+// a bare `PATH=/usr/bin:/bin:/usr/sbin:/sbin`. The `code` shim lives in /usr/local/bin
+// (and Homebrew casks use /opt/homebrew/bin), neither of which is on that PATH, so
+// `command -v <executable>` always failed and the extension only ever showed
+// "Command not found". These are searched in addition to the inherited PATH.
+export const MACOS_EXTRA_PATHS = ["/usr/local/bin", "/opt/homebrew/bin"];
+
+// Last-resort lookup: the CLI shipped inside the application bundle itself, which
+// exists even when the user never ran "Shell Command: Install 'code' command in PATH".
+export const MACOS_APP_BUNDLES: Record<VSCodeFlavour, string> = {
+    [VSCodeFlavour.Code]: "Visual Studio Code",
+    [VSCodeFlavour.CodeOSS]: "Code - OSS",
+    [VSCodeFlavour.Cursor]: "Cursor",
+    [VSCodeFlavour.VSCodium]: "VSCodium",
+    [VSCodeFlavour.Antigravity]: "Antigravity",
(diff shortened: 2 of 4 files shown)


clankus-aurelius:

clankus-aurelius (approved):
No actionable issues found in the supplied changes.

---

clankus-aurelius:
@aurelleb automated review passed for `f5d87c0`; this extension is ready for your review.

PR #369, opened 2026-08-30
Add JWT extension

balazsorban44:
Decode a JSON Web Token from the clipboard and read it without leaving the keyboard.

## Commands

**Decode JWT** takes the token from the clipboard, or as a command argument, and presents it as sections you move between with the arrow keys. Enter acts on the selected one.

| Section | Enter | Shows |
|---|---|---|
| Payload | Copy JSON | Claims annotated with their registered names, timestamps as dates |
| Header | Copy JSON | The JOSE header, annotated the same way |
| Signature | Verify | The verdict, the algorithm, the issuer's discovery document |
| Diff | Copy patch | A unified diff against the previously decoded token |
| Specs | Open RFC 7519 | RFC 7519, RFC 7515, RFC 9068, OpenID Connect Core |

Claim names come from those same four specs. Unregistered claims are left alone.

**JWT Expiry** is a no-view command that reads the clipboard and shows a HUD: `Active - Aug 30, 2026, 2:52 AM - in 2 hours - web-app - 1234567890`.

## Verification

Verification never runs on its own, only when Enter is pressed on the Signature row, and it only contacts the token's own issuer.

- **RS, ES, PS**: the `iss` claim leads to the discovery document, then the JWKS, then the key matching `kid`.
- **HS256/384/512**: a form asks for the shared secret. It is tried as raw text and as base64url, the result reports which matched, and it is never written to disk.

Verdicts are `Successful`, `Failed`, or `Unknown` when the check could not be completed (no `iss`, unreachable JWKS, or a `kid` the issuer does not publish). `Unknown` is deliberately distinct from `Failed`.

## Storage

The previously decoded token is kept in `LocalStorage` so the next one can be diffed against it. Only the header and payload claims are stored, never the signature, so nothing at rest can be replayed as a credential. A `Forget Previous Token` action clears it.

## Notes

- No dependencies beyond `@vicinae/api` and `jose`.
- 34 unit tests cover decoding, the claim dictionary, the diff, the annotation and the HMAC verification paths. Run them with `node --test src/*.test.ts`.
- Source and issue tracker: https://github.com/balazsorban44/vicinae-jwt


clankus-aurelius:

extensions/jwt/src/annotate.ts
```diff
@@ -0,0 +1,29 @@
...
+  const annotation = parts.filter(Boolean).join(" · ");
+  return annotation || null;
+}
+
+/**
+ * The decoded section as JSON with each registered claim explained on its own line.
+ * Not valid JSON any more: it is read, not parsed. Copy actions use the plain form.
+ */
+export function annotate(claims: Claims): string {
+  const lines = JSON.stringify(claims, orderClaims(Object.keys(claims)), 2).split("\n");
```
clankus-aurelius: **🔴 Blocking — Nested claim fields disappear from the decoded view**

Rule: `CORRECTNESS-001`

Passing the ordered top-level keys as JSON.stringify's replacer array also filters every nested object. Claims such as `address`, `act`, or custom object claims therefore appear empty or incomplete.

**Suggested resolution:** Order the top-level object before serialization, then stringify it without a replacer array so nested fields remain intact.

```suggestion
  const ordered = Object.fromEntries(orderClaims(Object.keys(claims)).map((key) => [key, claims[key]]));
  const lines = JSON.stringify(ordered, null, 2).split("\n");
```

extensions/jwt/src/decode.tsx
```diff
@@ -0,0 +1,406 @@
...
+/** Action panel order: ⏎ opens the first entry, the rest are one keypress further down. */
+const SPECS = [
+  { title: "RFC 7519: JSON Web Token", url: "https://www.rfc-editor.org/rfc/rfc7519" },
+  { title: "RFC 7515: JSON Web Signature", url: "https://www.rfc-editor.org/rfc/rfc7515" },
+  { title: "RFC 9068: JWT Access Tokens", url: "https://www.rfc-editor.org/rfc/rfc9068" },
+  { title: "OpenID Connect Core", url: "https://openid.net/specs/openid-connect-core-1_0.html" },
+] as const;
+
+/** Same claim order as the annotated view, so `iss`/`sub`/`exp` lead instead of whatever sorts first. */
+const pretty = (claims: Claims) => JSON.stringify(claims, orderClaims(Object.keys(claims)), 2);
```
clankus-aurelius: **🔴 Blocking — Copied payloads and diffs omit nested data**

Rule: `CORRECTNESS-001`

The replacer array applies recursively, so `pretty()` removes nested keys. Copy Payload can silently lose claims, and diffs can miss changes inside nested objects.

**Suggested resolution:** Construct an object with ordered top-level entries and stringify it without a replacer array.

```suggestion
const pretty = (claims: Claims) =>
  JSON.stringify(
    Object.fromEntries(orderClaims(Object.keys(claims)).map((key) => [key, claims[key]])),
    null,
    2,
  );
```

extensions/jwt/src/verify.ts
```diff
@@ -0,0 +1,92 @@
...
+  } catch {
+    return { state: "invalid" };
+  }
+}
+
+/**
+ * A shared secret is written down either as raw text or base64url, and the token gives no
+ * hint which. Both are tried, and the one that matched is reported back.
+ */
+export async function verifyWithSecret(token: string, secret: string): Promise<Verification> {
```
clankus-aurelius, line 118: **🔴 Blocking — Incomplete verification is reported as a failed signature**

Rule: `CORRECTNESS-001`

Every `compactVerify` error becomes `invalid`. For example, a JWKS with no selectable key—such as an empty set or multiple candidates when `kid` is absent—cannot complete verification but is presented as “Failed” instead of the documented “Unknown”.

**Suggested resolution:** Distinguish signature-verification failures from key-selection, unsupported-algorithm, and malformed-JWKS errors; map only an actual signature mismatch to `invalid` and return an unavailable or unknown result otherwise.

extensions/jwt/src/verify.ts
```diff
@@ -0,0 +1,92 @@
...
+    const discovery = await getJson<{ jwks_uri?: string }>(discoveryUrl(issuer));
+    if (!discovery.jwks_uri) {
+      return { state: "unavailable", reason: "Discovery document has no `jwks_uri`" };
+    }
+    jwks = await getJson<JSONWebKeySet>(discovery.jwks_uri);
+  } catch (error) {
+    return { state: "unavailable", reason: (error as Error).message };
+  }
+
+  if (kid && !jwks.keys.some((key) => key.kid === kid)) return { state: "unknown-key" };
```
clankus-aurelius, line 104: **🔴 Blocking — Signature keys may be retrieved over plaintext HTTP**

Rule: `NETWORK-001`

The issuer explicitly accepts `http://`, and the discovery document's `jwks_uri` is fetched without requiring HTTPS. An on-path attacker can replace the discovery response or signing keys and cause a forged token to be reported as successfully verified.

**Suggested resolution:** Require HTTPS for both the issuer discovery URL and `jwks_uri`, and prevent redirects from downgrading either request to HTTP.

clankus-aurelius (changes requested):
The extension has four publication-blocking correctness and network-safety issues.

---
_Automated review found 4 publication-blocking issues._

extensions/jwt/src/annotate.ts
balazsorban44: Confirmed and fixed in 208200c. Reproduced first:

```
JSON.stringify({sub:"u1",address:{street_address:"1 Main St"}}, ["sub","address"], 2)
// { "sub": "u1", "address": {} }
```

Rather than applying your suggestion in two places, the ordering now lives in a single `orderedJson` helper in `annotate.ts` that both the annotated view and the copy actions call, so the two cannot diverge again. Regression test: `orders top-level claims without filtering nested ones`, plus one asserting a nested `address` survives the annotated view.

extensions/jwt/src/decode.tsx
balazsorban44: Fixed in 208200c. `pretty` is now an alias for the shared `orderedJson` helper described in the `annotate.ts` thread, so Copy Payload, Copy Header and the diff all serialize through the same code path. The diff was affected too: it compares the pretty-printed forms, so nested changes were invisible.

extensions/jwt/src/verify.ts
balazsorban44: Fixed in 208200c. `verify.ts` now has a `classify` function that maps jose errors rather than collapsing them:

- `JWSSignatureVerificationFailed` is the only one that yields `invalid`
- `JWKSNoMatchingKey` yields `unknown-key`
- `JWKSMultipleMatchingKeys`, `JWKSInvalid` and `JWKInvalid` yield `unavailable` with a reason
- `JOSENotSupported` and `JOSEAlgNotAllowed` yield `unsupported`

`verifyWithSecret` had the same flaw and got the same treatment: it tries the secret as text and as base64url, and now only reports `invalid` if at least one attempt was a real signature mismatch. Regression test: `reports an empty key set as unknown rather than a failed signature`.

extensions/jwt/src/verify.ts
balazsorban44: Fixed in 208200c. Both requests now require https, and the response is rejected if its final URL is not https, which covers a redirect downgrade:

```ts
if (!url.startsWith("https://")) throw new Error(`${url} is not https`);
const response = await fetch(url, { signal: AbortSignal.timeout(5000) });
if (!response.url.startsWith("https://")) throw new Error(`${url} redirected off https`);
```

An http `iss` short circuits before any request is made. I had allowed http purely so a local test issuer could be used; that script served plaintext and is removed, with the JWKS path now covered by unit tests against a stubbed fetch. Three regression tests: http issuer, non-https `jwks_uri`, and a discovery response that redirected off https.

extensions/jwt/src/verify.ts
```diff
@@ -0,0 +1,127 @@
...
+ * attacker who can rewrite the discovery document or the key set can have a forged token
+ * report as verified. Redirects are followed but the final URL is checked too, so an
+ * https URL cannot be downgraded on the way.
+ */
+async function getJson<T>(url: string): Promise<T> {
+  if (!url.startsWith("https://")) throw new Error(`${url} is not https`);
+
+  const response = await fetch(url, { signal: AbortSignal.timeout(5000) });
+  if (!response.url.startsWith("https://")) throw new Error(`${url} redirected off https`);
+  if (!response.ok) throw new Error(`${url} responded ${response.status}`);
```
clankus-aurelius: **🔴 Blocking — Automatic redirects can still traverse plaintext HTTP**

Rule: `NETWORK-001`

`fetch` follows redirects automatically, while only the final `response.url` is checked. An HTTPS → HTTP → HTTPS chain therefore passes; an on-path attacker can tamper with the HTTP hop and redirect to an attacker-controlled HTTPS discovery document or JWKS, allowing forged tokens to appear verified.

**Suggested resolution:** Handle redirects manually and validate every redirect target as HTTPS before following it, for both discovery and JWKS requests.

clankus-aurelius (changes requested):
One prior publication blocker remains unresolved: redirect chains can still downgrade through HTTP.

---

extensions/jwt/src/verify.ts
balazsorban44: Right, the final-URL check was not enough. Fixed in d3ed14d: redirects are followed by hand instead, so every hop is validated before it is requested.

```ts
for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
  if (!target.startsWith("https://")) throw new Error(`${target} is not https`);
  const response = await fetch(target, { redirect: "manual", signal: AbortSignal.timeout(5000) });
  if (response.status >= 300 && response.status < 400) {
    const location = response.headers.get("location");
    if (!location) throw new Error(`${target} redirected without a location`);
    target = new URL(location, target).toString();
    continue;
  }
  ...
}
```

An http hop is now never requested at all, rather than rejected after the response comes back. Relative Location headers resolve against the current target, and the chain is capped at five hops.

Two regression tests: one asserts that a redirect to http is refused *and* that only the original https URL was ever fetched, the other that a redirect loop terminates.

clankus-aurelius (approved):
The prior redirect-downgrade blocker is resolved: redirects are now handled manually and every target is checked before it is requested. No new actionable issue was introduced by the incremental changes.

---

clankus-aurelius:
@aurelleb automated review passed for `d3ed14d`; this extension is ready for your review.

clankus-aurelius (approved):
The incremental README and screenshot changes introduce no actionable issues.

---

aurelleb:
simple and very useful, thank you!

PR #382, opened 2026-09-08
feat(omarchy-menu): load installed Omarchy 4 menu definitions at runtime

abelfubu:
## What

Replaces the statically typed Omarchy menu with runtime loading of the menu definitions installed by Omarchy 4, so the extension follows Omarchy updates deterministically instead of requiring code changes per release.

- Reads `/usr/share/omarchy/default/omarchy/omarchy-menu.jsonc` and merges `~/.config/omarchy/extensions/omarchy-menu.jsonc` over it in the same order as the Omarchy Shell menu (normalize each source, merge complete entries), including the synthetic-root fallback (`mergeMenuSources` parity).
- `$OMARCHY_PATH` is honored for the packaged paths, matching how Omarchy itself locates its files.
- JSONC stripping intentionally mirrors Omarchy's own `MenuModel.js` handling (no `jsonc-parser` dependency).
- Every parsed field is individually coerced, so malformed user entries cannot crash icon rendering or action handling; a malformed packaged menu degrades to the synthetic root instead of a fatal error.
- `when`/`checked` guards are evaluated through the installed `MenuModel.js` batcher (`guardScript`) with a compatibility fallback. Guard failures or timeouts keep the menu usable; a `when` item is hidden only on an explicit false, matching native behavior.
- Fonts and power profiles render as dynamic provider lists. Selecting a value keeps Vicinae open, awaits the setter, then revalidates so the checked indicator moves immediately.
- Shell-only providers (e.g. Apps) fall back to `omarchy menu summon <id>` so they keep their native behavior.
- Icons are converted from Nerd Font/Omarchy glyphs into tinted SVG data URIs; aliases are exposed as Vicinae keywords.

## Known limitation

Native Omarchy's launcher searches menu descendants globally from the root; this extension keeps its existing hierarchical navigation, with direct aliases available as keywords. Happy to follow up if global search parity is desired.

## Verification

- `tsc --noEmit`, `vici lint`, `vici build` pass.
- Runtime smoke tests on Omarchy 4: 320 menu items load in ~850 ms with correct `when`/`checked` state (7 checked rows match the system), 5 fonts listed with the current one checked, malformed user/packaged menus degrade gracefully.


clankus-aurelius:

extensions/omarchy-menu/src/menu.ts
```diff
@@ -0,0 +1,317 @@
...
+      const [index, value] = line.split(":");
+      results.set(Number(index), value === "1");
+    }
+  }
+
+  guards.forEach(({ item, kind }, index) => {
+    const result = results.get(index);
+    if (kind === "when" && result === false) item.parent = "";
+    if (kind === "checked" && result === true) item.isChecked = true;
+  });
```
clankus-aurelius, line 277: **🟠 Warning — Required menu parse failures are silently presented as an empty menu**

Rule: `UX-001`

`loadMenu` reads the required packaged menu through `readOptionalMenu`, whose catch converts malformed or unreadable content to `{}`. Because the preceding `access` only checks readability, a parse failure succeeds with a synthetic root and the UI reports “No available menu items” instead of identifying the broken installation.

**Suggested resolution:** Use a required-menu reader that propagates packaged-menu read and parse errors to the existing error view. If retaining the synthetic fallback is important, carry a warning into the UI rather than silently treating the failure as an empty menu.

extensions/omarchy-menu/src/menu.ts
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  let stdout: string;
+  try {
+    const result = await execFileAsync("/bin/bash", ["-lc", script], {
+      maxBuffer: 1024 * 1024,
+      timeout: 30_000,
+    });
+    stdout = result.stdout;
+  } catch {
```
clankus-aurelius: **🟠 Warning — Guard failures silently produce inaccurate menu state**

Rule: `UX-001`

Any guard batch error or 30-second timeout is swallowed, so conditional entries remain visible and checked entries appear unchecked without indicating that evaluation failed.

**Suggested resolution:** Keep the fallback state if desired, but propagate or return the guard failure so the command can show an actionable warning or error to the user.

clankus-aurelius (approved):
The runtime loader suppresses two material operational failure modes, leaving users with misleading menu state and no actionable explanation.

---

clankus-aurelius:
@aurelleb automated review passed for `25534f2`; this extension is ready for your review.

extensions/omarchy-menu/src/menu.ts
```diff
@@ -0,0 +1,381 @@
...
+      const [index, value] = line.split(":");
+      results.set(Number(index), value === "1");
+    }
+  }
+
+  guards.forEach(({ item, kind }, index) => {
+    const result = results.get(index);
+    if (kind === "when" && result === false) item.parent = "";
+    if (kind === "checked" && result === true) item.isChecked = true;
+  });
```
clankus-aurelius, line 277: **🟠 Warning — Required menu parse failures are silently presented as an empty menu**

Rule: `UX-001`

The required packaged menu is still read through `readOptionalMenu`, which catches read and parse errors and returns `{}`. A malformed readable installation therefore produces the synthetic root and “No available menu items” instead of explaining the failure.

**Suggested resolution:** Read the packaged menu with a required reader that propagates parse/read errors to `LoadError`, or carry an explicit warning into the UI when using the synthetic fallback.

extensions/omarchy-menu/src/menu.ts
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  let stdout: string;
+  try {
+    const result = await execFileAsync("/bin/bash", ["-lc", script], {
+      maxBuffer: 1024 * 1024,
+      timeout: 30_000,
+    });
+    stdout = result.stdout;
+  } catch {
```
clankus-aurelius: **🟠 Warning — Guard failures silently produce inaccurate menu state**

Rule: `UX-001`

Any guard batch error or 30-second timeout is still swallowed, leaving conditional entries visible and checked entries unchecked without telling the user evaluation failed.

**Suggested resolution:** Preserve the fallback state if desired, but return or propagate the guard failure so the command displays an actionable warning or error.

clankus-aurelius (approved):
Both previously reported operational failures remain unresolved: packaged-menu parse errors and guard-evaluation failures are still silently converted into misleading fallback state.

---

clankus-aurelius (approved):
The two previously reported UX issues are resolved by surfacing packaged-menu and guard-evaluation fallback warnings. No new actionable issues were introduced by the incremental changes.

---

abelfubu:
Addressed both UX-001 warnings in `3e9a95a`:

1. **Packaged-menu parse failures** — `loadMenu` now reads the required menu with I/O errors propagating to the error view (missing-install behavior unchanged), while a parse failure keeps the synthetic-root fallback but carries a concise warning ("The packaged Omarchy menu is malformed; only custom entries are shown") that is toasted once per command run. The warning text deliberately omits parser internals/paths.
2. **Guard failures** — a failed/timed-out guard batch still leaves all rows visible/unchecked (native `when` semantics), but now returns a warning ("Menu state checks failed/timed out; showing entries without them") that is toasted alongside any parse warning; both warnings are retained when they occur together.

Warnings are deduplicated per distinct message (Set-based), so re-renders and submenu pushes never re-toast. Verified end-to-end with simulated malformed packaged menu and broken `when` guards.

aurelleb:
thx!

PR #396, opened 2026-09-18
feat: add brave extension

actuallyaryaman:
includes extension for brave/origin browser


clankus-aurelius:

extensions/brave/src/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+	return BRAVE_VARIANTS.find((v) => v.id === variant)?.name ?? "Brave";
+}
+
+export function variantInfo(variant: BraveVariant): BraveVariantInfo {
+	return BRAVE_VARIANTS.find((v) => v.id === variant) ?? BRAVE_VARIANTS[0];
+}
+
+export function dataDirForVariant(variant: BraveVariant): string {
+	const custom = preferences.profile_dir?.trim();
+	return path.join(homedir(), custom ?? path.join(".config", variantInfo(variant).dataDir));
```
clankus-aurelius, line 54: **🟠 Warning — An empty profile preference resolves to the home directory**

Rule: `CORRECTNESS-001`

`trim()` produces `""`, but `custom ?? default` treats that as a configured path. Clearing the preference therefore searches the home directory instead of the documented per-browser default.

**Suggested resolution:** Treat an empty string as absent when selecting the profile directory.

```suggestion
	return path.join(homedir(), custom || path.join(".config", variantInfo(variant).dataDir));
```

extensions/brave/src/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return undefined;
+}
+
+export function isVariantInstalled(variant: BraveVariant): boolean {
+	const dir = dataDirForVariant(variant);
+	if (existsSync(path.join(dir, "Bookmarks")) || existsSync(path.join(dir, "History"))) return true;
+	return variantInfo(variant).binaries.some((b) => findOnPath(b) !== undefined);
+}
+
```
clankus-aurelius, line 74: **🟠 Warning — Profile-only installations are never detected**

Rule: `CORRECTNESS-001`

The installation check looks for `Bookmarks` and `History` at the user-data root, while `getBraveProfiles` correctly expects them inside profile subdirectories. Bookmark and history commands consequently reject an existing installation whenever its executable is not on `PATH`, despite the README requiring `PATH` only for window and tab commands.

**Suggested resolution:** Detect profile data by safely scanning the user-data directory's profile subdirectories, using the same criteria as `getBraveProfiles`.

extensions/brave/src/search-tabs.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const enableAndRestart = async () => {
+		try {
+			const flagsPath = ensureDebuggingFlag(variant);
+			const restarted = await quitBrave(variant);
+			launchBrave(variant, restarted ? ["--restore-last-session"] : []);
+			await closeMainWindow();
+			await showToast({
+				style: Toast.Style.Success,
+				title: "Debugging enabled",
+				message: `Flag written to ${flagsPath}; ${name} is restarting`,
```
clankus-aurelius, line 134: **🟠 Warning — Failed browser termination is reported as a successful restart**

Rule: `CORRECTNESS-001`

`quitBrave` returns `false` when the running browser cannot be identified, terminated, or stopped within ten seconds, but this result is ignored: the code launches Brave and reports that it is restarting. A still-running instance ignores the newly written startup flag, so Search Tabs remains unavailable despite the success message.

**Suggested resolution:** Record whether Brave was running and treat a false termination result as a failure; only launch and report success after the old process stopped, or when no process was running.

clankus-aurelius (approved):
The extension has three profile/debugging logic defects and includes an unauditable executable asset without provenance.

### Findings without an inline diff location

- **Vendored executable WASM has no established provenance** (`extensions/brave/assets/sql-wasm.wasm:1`, ASSET-001): The PR adds a 658 KB compiled WebAssembly executable that is loaded by `sql.js`, but supplies no source, reproducible-generation information, or evidence tying this blob to the integrity-pinned dependency, so it cannot be meaningfully audited. Establish the asset's exact upstream source and version with a reproducible verification step, or generate/copy it from the pinned `sql.js` package instead of committing an unexplained binary.

---

clankus-aurelius:
@aurelleb automated review passed for `7ca1c3b`; this extension is ready for your review.

PR #398, opened 2026-09-19
feat: add kcd-connect extension

dagimg-dot:
## KCD Connect

Control a paired phone from the launcher through [**kcd**](https://github.com/bethropolis/kcd), a headless KDE Connect daemon.

The extension talks to kcd's Unix socket at `$XDG_RUNTIME_DIR/kcd/kcd.sock`. It spawns **no processes** — no `kdeconnect-cli`, no KDE dependencies — and makes no network requests. One socket connection carries a full snapshot of every known device plus a live event stream, so nothing polls.

### Commands

| Command | What it does |
| :-- | :-- |
| **Devices** | Live overview of every paired device with battery, signal, now-playing and a full action panel |
| **Send File** | Send files with real byte-level transfer progress |
| **Send Clipboard** | Push the clipboard to a device |
| **Ring Device** | Play a ringtone to locate a device |
| **Send SMS** | Compose a message, with contacts synced from the phone |
| **Media Control** | Play/pause, skip and seek what the phone is playing |
| **Pair Device** | Discover and pair, with verification-key confirmation |

### Screenshots

Devices

Device actions

Media Control

Send SMS

### Two design notes

**Transfer progress is event-driven.** kcd's `share` returns as soon as the invite packet is queued — the bytes move afterwards in a background goroutine. Awaiting it and reporting "sent" would claim success before anything was transferred, so progress and completion come from the `share.progress` / `share.complete` events instead.

**Multi-device without the prompting.** Actions resolve a target through a ladder — explicit choice, pinned default, last used, sole connected device — and only ask when the answer is genuinely ambiguous. One device behaves like a single-device tool; several stop nagging. Offline devices stay listed with their last known battery rather than being hidden, and broadcast reports partial failure honestly instead of rounding up to success.

### Not included

Three things are deliberately absent because the daemon cannot support them correctly, documented in the README:

- **Send Text** — kcd can receive shared text but exposes no route to send it; its share command takes a file path, and `clipboard_push` reads the real system clipboard.
- **Run Command** — the remote-command list request is forwarded to the phone, but the reply has no event on the daemon's bus, so a client cannot read the list back.
- **SMS conversation browsing** — sending works; reading threads needs correlation work beyond this scope.

### Requirements

kcd running (`systemctl --user enable --now kcd`) and a device paired via the standard KDE Connect app.

### Checks

`make validate EXT=kcd-connect`, `npm run lint`, `npm run check` (Biome), `tsc --noEmit` and `npm run build` all pass. Verified against a real paired phone: device state and live events, media control, and a file transfer completing end to end.


clankus-aurelius:

extensions/kcd-connect/src/send-file.tsx
```diff
@@ -0,0 +1,299 @@
...
+								error: p.error ?? "transfer failed",
+							});
+						}
+						if (pending.size === 0) finish();
+						else resetIdle();
+					}
+				},
+
+				onError: () => finish(),
+				onClose: () => finish(),
```
clankus-aurelius: **🔴 Blocking — Event-stream failures are reported as successful transfers**

Rule: `CORRECTNESS-001`

Both watch failure callbacks call `finish()` without recording a failure. If the stream fails before readiness, `pending` is empty and every transfer is counted as sent; if it fails during transfer, `failures` still remains empty, so `handleSubmit` displays a success toast despite missing completion events.

**Suggested resolution:** Reject the transfer on watch errors or convert every unconfirmed transfer into a failure, and only show success when every expected `share.complete` event was received.

extensions/kcd-connect/src/send-file.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+		const watch = openWatch(
+			[EventType.ShareProgress, EventType.ShareComplete],
+			{
+				onReady: () => {
+					void (async () => {
+						for (const device of targets) {
+							for (const file of files) {
+								const base = path.basename(file);
+								pending.add(key(device.id, base));
```
clankus-aurelius: **🔴 Blocking — Files with the same basename share one completion key**

Rule: `CORRECTNESS-001`

Pending transfers are stored in a `Set` keyed by device ID and `path.basename(file)`. Selecting two files with the same basename for one device collapses them into one entry, so the first completion closes the watcher and the command claims both files succeeded.

**Suggested resolution:** Reject duplicate basenames per target or track an occurrence count/unique transfer identifier that can be correlated with completion events.

extensions/kcd-connect/src/pair-device.tsx
```diff
@@ -0,0 +1,271 @@
...
+											content={device.id}
+											icon={Icon.CopyClipboard}
+										/>
+									</ActionPanel.Section>
+									{globalActions}
+								</ActionPanel>
+							}
+						/>
+					))}
+				</List.Section>
```
clankus-aurelius, line 289: **🔴 Blocking — Incoming pairing can bypass verification-key confirmation**

Rule: `MANIFEST-001`

For `PAIR_REQUESTED_BY_PEER`, the action is presented as “Accept Pairing” and directly invokes `requestPairing`, which calls `pairDevice` without showing `ConfirmPairing`. This contradicts the advertised verification-key confirmation and bypasses it in a normal incoming-request path.

**Suggested resolution:** Route incoming requests through the verification screen and require the user to compare the key before invoking `pairDevice`; otherwise remove the verification claim and clearly disclose the limitation.

extensions/kcd-connect/src/components/DeviceListItem.tsx
```diff
@@ -0,0 +1,399 @@
...
+										})
+									}
+								/>
+							</ActionPanel.Section>
+
+							<ActionPanel.Section title="Device">
+								<Action
+									title="Ring"
+									icon={Icon.PhoneRinging}
+									onAction={() =>
```
clankus-aurelius, line 278: **🟠 Warning — Device actions do not preserve the selected target**

Rule: `CORRECTNESS-001`

“Send File…” and “Send SMS…” are launched from a specific device’s action panel without passing that device ID. The destination commands independently resolve a default, last-used, or first connected device, so they can open targeting a different device.

**Suggested resolution:** Pass the selected device ID through the command launch context or arguments and initialize each destination command from that explicit ID.

clankus-aurelius (changes requested):
The file-transfer flow can report success without confirmed completion, and pairing can bypass the advertised verification step. Device-scoped actions also fail to preserve the selected device.

---
_Automated review found 3 publication-blocking issues._

extensions/kcd-connect/src/send-file.tsx
```diff
@@ -0,0 +1,329 @@
...
+								await shareFile(job.deviceId, job.path);
+								const k = key(job.deviceId, job.base);
+								outstanding.set(k, [...(outstanding.get(k) ?? []), job]);
+							} catch (error) {
+								failures.push({
+									device: job.deviceName,
+									file: job.base,
+									error: error instanceof Error ? error.message : String(error),
+								});
+							}
```
clankus-aurelius, line 129: **🔴 Blocking — In-flight jobs are untracked and can be reported as sent**

Rule: `CORRECTNESS-001`

Each job is removed from `undispatched` before `shareFile` is awaited, but is added to `outstanding` only afterward. If an earlier completion empties `outstanding` during that await, `finish()` resolves every job as sent; a stream failure in the same window also omits the current job from `abort()`'s stranded set.

**Suggested resolution:** Register each job before awaiting `shareFile`, remove it on request failure, and allow completion only after dispatch has finished and no jobs remain outstanding.

extensions/kcd-connect/src/pair-device.tsx
```diff
@@ -0,0 +1,309 @@
...
+		<Detail
+			markdown={markdown}
+			actions={
+				<ActionPanel>
+					<Action
+						title={
+							secret
+								? `${kind === "key" ? "Keys" : "Fingerprints"} Match, Pair`
+								: "Pair Anyway"
+						}
```
clankus-aurelius: **🔴 Blocking — Pairing still permits unverified acceptance**

Rule: `MANIFEST-001`

When neither a verification key nor fingerprint is supplied, the action becomes “Pair Anyway” and invokes `pairDevice`. This still bypasses verification despite the README and PR description promising verification-key confirmation.

**Suggested resolution:** Disable pairing when no verifiable value is available, or remove the verification-key claim and clearly disclose that unverified pairing is permitted.

clankus-aurelius (changes requested):
Two publication-blocking issues remain: file transfers can still be counted as complete prematurely, and pairing still permits bypassing the advertised verification step.

---
_Automated review found 2 publication-blocking issues._

clankus-aurelius (approved):
The two prior blocking findings are resolved by the incremental changes. No new actionable issue was introduced.

---

clankus-aurelius:
@aurelleb automated review passed for `05669b8`; this extension is ready for your review.

PR #399, opened 2026-09-19
feat(niri): add a command for browsing configured keybinds

dagimg-dot:
## Keybinds

Adds a `Keybinds` view command to the niri extension: a searchable list of every
bind declared in the niri config, grouped by what the action targets
(Applications, Screenshots, Monitors, Workspaces, Columns, Windows, Compositor).

niri exposes no IPC request for binds — `niri msg` has no `binds` subcommand — so
the command reads the config file rather than asking the compositor:

- `$NIRI_CONFIG`, else `$XDG_CONFIG_HOME/niri/config.kdl`, else `/etc/niri/config.kdl`
- `include` nodes are followed, resolved relative to the file that declares them,
  with `~` expanded and a guard against cycles
- multiple `binds` blocks across files are merged, as niri does
- an include that cannot be read is skipped rather than failing the command:
  that is niri's error to report, and listing the binds we could read beats
  listing none

niri 26.04 parses **KDL v1**, not v2 — raw strings are `r#"..."#`, and `#"..."#`
is a syntax error. `src/lib/keybinds/kdl.ts` is a small reader for the subset
niri accepts: nested block comments, `/-` node comments, escline, quoted and raw
strings, type annotations. It is forgiving on purpose, for the same reason as
above.

Each item shows the key combination, and the bind's `hotkey-overlay-title` when
it has one, falling back to the action and its arguments. Bind properties surface
as tags — `No Repeat`, `150ms Cooldown`, `Works When Locked`, `Never Inhibited`,
and `Hidden From Overlay` for `hotkey-overlay-title=null`. The detail pane adds
the action, its arguments, and the file and line the bind was declared in, which
starts to matter once `include` is in play. Actions cover copying the keys, the
command, the bind's own source text and the config path, plus a reload.

## Also in this PR

- `refactor(niri)`: the root of `src/` now holds one file per manifest command
  and nothing else. Shared code moved under `components/` for UI and `lib/` for
  everything else, with `lib/keybinds/` for the config reader. `utils.ts` split
  along the seam it already had: `lib/niri.ts` runs `niri msg`, `lib/feedback.ts`
  shows toasts.
- `style(niri)`: four files failed `npm run format:check` on a clean checkout.
  Whitespace only.

## Testing

Parsing was checked against a real config — 128 binds, with `//` and `/-`
commented binds excluded and multi-line binds kept intact — and against a
synthetic file covering raw strings, escaped quotes, nested block comments,
escline, action property arguments (`move-column-to-workspace 2 focus=false`),
`hotkey-overlay-title=null` and multiple `binds` blocks; plus a nested, cyclic
and missing-include tree. Every claim about niri's syntax above was confirmed
with `niri validate` on niri 26.04 rather than taken from the docs.

The new command and the refactored ones were run in Vicinae after the move.
`npm run lint`, `npm run format:check` and `make validate EXT=niri` pass, and
`npm run typecheck` is clean for every file this PR touches.

## Known, not addressed here

`npm run typecheck` is *not* clean for the extension as a whole: the other
commands reference `Icon.Window`, `Icon.Gear` and `Icon.ExclamationMark`, none of
which exist in `@vicinae/api`. That is 24 pre-existing errors, and since
`vici build` does not typecheck, those commands ship with `undefined` icons and
render nothing where an icon should be. Left alone to keep this PR to one
concern — happy to fix it in a follow-up if you'd like it here instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


clankus-aurelius:

extensions/niri/src/lib/keybinds/config.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (node.name !== 'include') continue;
+
+    const target = node.args[0];
+    if (typeof target !== 'string') continue;
+
+    try {
+      await readConfigTree(resolveIncludePath(target, absolute), seen, files);
+    } catch {
+      // An include niri cannot read is niri's to report: listing the binds we
+      // did read beats refusing to list any.
```
clankus-aurelius, line 52: **🟠 Warning — Unreadable includes are silently omitted**

Rule: `UX-001`

The catch suppresses every include read failure, so the command appears successful while configured keybinds may be missing. No toast or in-view warning tells the user that the displayed list is incomplete.

**Suggested resolution:** Continue loading readable files, but collect failed include paths and show actionable feedback indicating that some keybinds could not be loaded.

clankus-aurelius (approved):
The new keybind browser can silently present an incomplete list when an included config file cannot be read.

---

clankus-aurelius:
@aurelleb automated review passed for `6688a62`; this extension is ready for your review.

